PR #3, opened 2018-10-16
Make nick in GuildMemberUpdate a Maybe

PixeLInc:
The `nick` field in `GuildMemberUpdate` is actually a nullable field and I was getting some errors about users without a nickname.
This just sets the type to `Maybe` so it can be nulled out.

Please let me know if I need to make changes other places, though.

### DIFF
--- a/src/Discord/Types/Events.hs
+++ b/src/Discord/Types/Events.hs
@@ -36,7 +36,7 @@ data Event =
   | GuildIntegrationsUpdate Snowflake
   | GuildMemberAdd          Snowflake GuildMember
   | GuildMemberRemove       Snowflake User
-  | GuildMemberUpdate       Snowflake [Snowflake] User String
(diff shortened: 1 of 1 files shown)


aquarial:
Thanks for field testing the library!

You should also change the parsing. Change the `.:` where `GuildMemberUpdate` is parsed into a `.:?`

```
~> :type (.:)
(.:) :: FromJSON a => Object -> T.Text -> Parser a
~> :type (.:?)
(.:?) :: FromJSON a => Object -> T.Text -> Parser (Maybe a)
```

`.:` will work if the field is present or "null" but fail if the field isn't present. `.:?` will parse `Nothing` if the field isn't present. Slightly more general.

PixeLInc:
There we go, thanks :)

aquarial:
Looks good

Issue #1, opened 2018-08-30
Thread blocked indefinetly in an MVar operation


PixeLInc:
^ Title
Using GHC: 8.4.3

I get the error in the title when trying to run the example codes.

Code:
```haskell
{-# LANGUAGE OverloadedStrings #-}
module Main where

import Control.Monad
import Control.Exception
import Data.Char (isSpace, toLower)
import qualified Data.Text as T
import qualified Data.Text.IO as TIO

import Discord

main :: IO ()
main = do
    tok <- T.filter (not . isSpace) <$> TIO.readFile "auth-token.secret"
    dis <- loginRestGateway (Auth tok)

    finally (forever $ do
                e <- nextEvent dis
                case e of
                    MessageCreate m -> when (isPing (messageText m)) $ do
                        resp <- restCall dis (CreateMessage (messageChannel m) "Pong!" Nothing)
                        putStrLn $ show resp

                    _ -> pure ())
            (stopDiscord dis)

isPing :: T.Text -> Bool
isPing = T.isPrefixOf "ping" . T.map toLower
```

I'm slowly learning Haskell so I don't really know what's going on other than it's probably in the lib or something.
Lmk if there's any more info you need or files (like the stack or cabal)

mattgreen:
I'm getting this as well with a valid auth token.

mattgreen:
I figured it out! I was putting in the client secret for my app instead of the bot token. The bot token is listed under the bot section of your Discord app settings. I should know better, given that I already have a live bot.

The library needs to funnel back authentication errors instead of writing them to a log file, but hopefully this gets you unstuck.

PixeLInc:
That's the thing, I'm 100% sure I'm using a valid auth token as well. It works on other bots. I'll look more into it I guess.

PixeLInc:
Well, shit. I feel kinda stupid now.
I was prefixing my token with 'Bot' like how you normally would and it wasn't working but removing that lets it work fine.

Derp. Thanks for the tip.
(But also it should tell me the token is wrong instead of breaking but.. heh)

aquarial:
I'm definitely less active than over the summer.

I just fixed the "Bot" prefix incompatibility by [only adding the prefix if it's not present](https://github.com/aquarial/discord-haskell/blob/974c3afad4c7e30db18d9421907fbebb9b2d6f59/src/Discord/Types/Prelude.hs#L26)

[...]

PR #7, opened 2019-01-15
Handle CloseRequest 1001 event from Discord gateway

t1m0thyj:
This should fix the following error which I encountered (#1):

> Event error: GatewayExceptionConnection (CloseRequest 1001 "") "Normal event loop close request"
> discord-haskell-bot-exe: thread blocked indefinitely in an MVar operation

Not sure if this is the right way to fix it or kind of a hack. The [Discord API docs](https://discordapp.com/developers/docs/topics/opcodes-and-status-codes#rpc) don't mention the error code 1001, but there is a list of close event codes on MDN [here](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent). Seeing the list there makes me wonder, should other codes be handled too, like 1002, 1003, 1007, etc.?

I should also mention that I haven't gotten to test this yet, can run my bot for a week with this PR to verify that my fix prevents it from crashing.

EDIT: My guess would be that it would be best to not handle other 100x error codes unless they come up in the future. 1001 seems like a special case with Discord websockets that happens because of Cloudflare.

### DIFF
--- a/src/Discord/Gateway/EventLoop.hs
+++ b/src/Discord/Gateway/EventLoop.hs
@@ -174,8 +174,9 @@ eventStream (ConnData conn seshID auth eventChan) seqKey interval send log = loo
     eitherPayload <- getPayloadTimeout conn interval log
     case eitherPayload :: Either ConnectionException GatewayReceivable of
       Left (CloseRequest code str) -> case code of
-          -- see discord documentation on gateway close event codes
+          -- see Discord and MDN documentation on gateway close event codes
           1000 -> ConnReconnect auth seshID <$> readIORef seqKey
+          1001 -> ConnReconnect auth seshID <$> readIORef seqKey
           4000 -> ConnReconnect auth seshID <$> readIORef seqKey
           4006 -> pure ConnStart
           4007 -> ConnReconnect auth seshID <$> readIORef seqKey


aquarial:
The discord docs don't say what they expect a bot to do with the various event codes. A `4004 - Invalid Authentication` shouldn't reconnect, but some of the others aren't clear. 

There also isn't any consensus in the libraries I can see. The python disco library [doesn't resume for 4000 to 4010 codes](https://github.com/b1naryth1ef/disco/blob/1b3c7fb7a61352a48f2c9d39f0436e06a848d808/disco/gateway/client.py#L253), discord.js [doesn't resume](https://github.com/discordjs/discord.js/blob/21ec03140f182b93ddc151e9edbdb6da8d8ae703/src/client/websocket/WebSocketShard.js#L347) for [1000, 4004, 4010, 4011 codes](https://github.com/discordjs/discord.js/blob/8230255c68b94d68a4e8ffc559a98d08d1a08a7c/src/util/Constants.js#L91), and java Discord4J has an [if-conditional that checks a few other 40XX codes](https://github.com/Discord4J/Discord4J/blob/be2318cf518cb97bf02cbabef0695b15ea3e5953/src/main/java/sx/blah/discord/api/internal/DiscordWS.java#L198).  

This change will fix the specific issue of a `1001` disconnect, so all should be good for now.

PR #11, opened 2019-02-20
GuildRoleDelete: returns role_id instead of role

MP2E:
While running my bot, I noticed another crash that happened whenever a Role was deleted.

```
Event error: GatewayExceptionEventParseError "Error in $: key \"role\" not present"
"Normal event loop"
```

From reading through the Discord documentation on Roles, it seems that GuildRoleDelete returns a role_id instead of a full role object.

https://discordapp.com/developers/docs/topics/gateway#guild-role-delete

I think I've made the right changes here, hope this helps!


aquarial:
Yup! Thanks for the fix

PR #18, opened 2019-05-05
Follow more closely the Embed API

marcotoniut:
Support for optional values and more accurate field names.
[Embed Documentation](https://discordapp.com/developers/docs/resources/channel#embed-object-embed-footer-structure)


aquarial:
Thanks, looks good

PR #19, opened 2019-05-08
Simply ping-pong.hs

cnguy:
Hi, thanks for the awesome package.

Adding a quickfix to `examples/ping-pong.hs`.

Changelog states (for latest aka v0.8.3): **Simplify Message Author from `Either WebhookId User` to `User`**

So `fromBot` can be simplified.


aquarial:
Nice

PR #35, opened 2020-03-26
Implement Read instance for Snowflake

rexim:
I'm not sure why this instance was not implemented, but I really need it to simplify some of the serialization I do in my bot that uses your beautiful library (thank you so much for implementing it <3).

Please let me know if there are any problems with my PR. Thanks!

### DIFF
--- a/src/Discord/Internal/Types/Prelude.hs
+++ b/src/Discord/Internal/Types/Prelude.hs
@@ -14,8 +14,8 @@ import Data.Time.Clock.POSIX
 import Data.Monoid ((<>))
 import Control.Monad (mzero)
 
-import Data.Functor.Compose
-import Data.Bifunctor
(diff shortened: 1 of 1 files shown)


src/Discord/Internal/Types/Prelude.hs
```diff
@@ -14,6 +14,9 @@ import Data.Time.Clock.POSIX
 import Data.Monoid ((<>))
 import Control.Monad (mzero)
 
+import Data.Functor.Compose
+import Data.Bifunctor
+
 -- | Authorization token for the Discord API
```
aquarial: I prefer explicit import lists, especially for less common imports like these,

aquarial:
I'm going to make sure the project will build on stackage lts-15.5, and then push this change as version 1.5.0 of the library.

PR #37, opened 2020-04-07
Added Response type family and caching of some requests

### LINKED ISSUE #36: Proposal: Cache User Requests
I am writing a chat bot that uses DMs as its main interaction. I was facing the problem that a message sent to a bot from a DM was not in the cache's DM channels map. As a result I had to write this snippet of code
```haskell
isDM :: ChannelId -> DiscordHandle -> IO Bool
isDM ch dh = do
  c <- readCache dh
  case M.lookup ch (_dmChannels c) of
    Just chan ->
      if isDMChannel chan
        then pure True
        else pure False
    _ ->
      putStrLn "not in cache" >>
      (isDMChannel <$> (fromRight' <$> restCall dh (R.GetChannel ch)))
  where
    isDMChannel (ChannelDirectMessage {}) = True
    isDMChannel _ = False
```
Testing my code results in the cache missing every time requires a call to Discord to query the same channel id.

I propose that we add a `updateCache` function to the Request typeclass. The default implemention would be `id`, but requests that can be cached save their result.

What do you think about this proposal?

rand0m-cloud:
Closes #36.

It is incomplete for most requests, right now only the GetChannel will update the cache.

I do not have the same formatter as you, so you have to re-format my files (sorry).

It compiles without any new warnings and I don't believe it will require any testing.

It's my first contributing to a Haskell package, so any suggestions are welcome.

### DIFF
--- a/src/Discord.hs
+++ b/src/Discord.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
@@ -117,14 +118,18 @@ data RestCallErrorCode = RestCallErrorCode Int T.Text T.Text
   deriving (Show, Eq, Ord)
 
 -- | Execute one http request and get a response
-restCall :: (FromJSON a, Request (r a)) => DiscordHandle -> r a -> IO (Either RestCallErrorCode a)
+restCall :: (FromJSON (Response r), Request r) => DiscordHandle -> r -> IO (Either RestCallErrorCode (Response r))
 restCall h r = do empty <- isEmptyMVar (discordHandleLibraryError h)
                   if not empty
                   then pure (Left (RestCallErrorCode 400 "Library Stopped Working" ""))
                   else do
                       resp <- writeRestCall (discordHandleRestChan h) r
                       case resp of
-                        Right x -> pure (Right x)
+                        Right x -> do
+                            c <- readCache h
+                            let c' = updateCache c r x
+                            putMVar (snd $ discordHandleCache h) (Right c')
+                            pure (Right x)
                         Left (RestCallInternalErrorCode c e1 e2) ->
                           pure (Left (RestCallErrorCode c (TE.decodeUtf8 e1) (TE.decodeUtf8 e2)))
                         Left (RestCallInternalHttpException _) ->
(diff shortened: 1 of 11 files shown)


aquarial:
It's an interesting change. I'm not sure about it right now. 

I don't know what these things do. I've avoided a lot of advanced haskell types tbh.

```haskell
{-# LANGUAGE FlexibleContexts  #-}
{-# LANGUAGE AllowAmbiguousTypes  #-}
```

winadam:
Let me explain the problem and the reasons I made the change. To add a function to the `Request` typeclass that will take the resulting value of a request requires you add an associated type (`TypeFamilies`) `Response` that takes a request and turns it into the response value of that request. You were capturing this idea by doing `Request (r a)` and using the `a`. 

A problem arises when converting the signatures to use the new `Response a` type. `restCall` needs the response value to be decodable from json and `FlexibleContexts` was required because now the `FromJSON` constraint is placed on the response type, instead any variable `a`.

The `AllowAmbiguousTypes` extension was left in by accident and is not required.


Testing the new branch gave me weird symptoms that I am not able to consistently reproduce. In my application, I was sending a `GetChannel` rest call and noticing the application "freezing" until another message (or event?) was received. On some occasions the application would fail (or freeze) on creating a message. I am not sure if the issue lies in my code or the library. I want to do more investigation before I feel confident in letting this PR being merged.

Do you have any insight to offer on this issue? I am not that intimate with your codebase, am I modifying the cache in a erroneous manner?

src/Discord.hs
```diff
@@ -117,14 +118,18 @@ data RestCallErrorCode = RestCallErrorCode Int T.Text T.Text
...
                   if not empty
                   then pure (Left (RestCallErrorCode 400 "Library Stopped Working" ""))
                   else do
                       resp <- writeRestCall (discordHandleRestChan h) r
                       case resp of
-                        Right x -> pure (Right x)
+                        Right x -> do
+                            c <- readCache h
+                            let c' = updateCache c r x
+                            putMVar (snd $ discordHandleCache h) (Right c')
```
aquarial, line 131: This code is the cause of the slowness/crashes. The cache is stored in an `MVar`, which is a thread communication device which can store a single object. `readCache` calls `readMVar`, which is going to return a copy of the cache. The `MVar` remains full. Then `putMVar` is going to try and put `c` into the `MVar`, and will block until the `MVar` is empty. 

The cache is kept inside an `MVar` so many threads can safely modify it at once. Get the cache with `takeMVar`, modify it, then put it back with `putMVar`.

annogram: Have you tried switching to STM? It handles concurrency quite well and isn't a bit leap from MVars

rand0m-cloud:
Sorry to close this, but I have moved on to different projects and will not be able to commit any time to this issue.

PR #39, opened 2020-05-11
Add the ability to serialize snowflakes

annogram:
I've done this on my project manually for more efficient storage (was doing using json before)


aquarial:
This seems like a real use-case, but I'm not sure it has that much value to other people.

It's a pretty specific use, only implemented for snowflakes. They're just `Word64`. Even manually serializing to a `ByteString` seems reasonable

Or they can add the instance and deal with the orphan instance. I dunno. 

```haskell

import Data.Serialize

instance Serialize Snowflake where
  put = put . show
  get = read <$> get

```

It's a reasonable request, i'm just not into it

PR #47, opened 2020-07-19
Get channelName for categories

stridervc:
Get the names of channels that are of type category


aquarial:
Cool stuff plus the T.toLower is neat

I'll update hackage on Sunday

PR #49, opened 2020-07-21
Use Functor instance for Maybe instead of a case-expression

elikoga:


aquarial:
`<&>` is neat, makes sense

PR #50, opened 2020-07-21
Fix some warnings

elikoga:
Removes redundant imports issuing `warning: [-Wunused-imports]` on compile time

Adds type signatures to literals issuing `error: [-Wtype-defaults, -Werror=type-defaults]` on compile time when compiled with `stack build --pedantic`

### DIFF
--- a/examples/cache.hs
+++ b/examples/cache.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
 
-import Data.Monoid ((<>))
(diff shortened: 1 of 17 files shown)


aquarial:
The `import Data.Monoid ((<>))` let's the project build on an older version of base that doesn't import it by default.

I get home Sunday or so, and I'll take a look to see if that base is so out of date it's just worth dropping, or we should use compiler pragma guards or something.

I agree the pile of warnings isn't nice.

PR #84, opened 2021-12-17
Application commands

L0neGamer:
The neccessities for application commands.

Interactions need to be expanded to work with components (which don't exist currently), and this entire thing could really do with some more concrete and locked down implementations.

Has been tested with slash commands alone using ping pong.

Thanks to https://github.com/social-anthrax for getting this started.

### DIFF
--- a/examples/ping-pong.hs
+++ b/examples/ping-pong.hs
@@ -13,8 +13,7 @@ import Discord.Types
 import qualified Discord.Requests as R
 import Debug.Trace
 import Data.Maybe (fromJust)
-import qualified Discord.Internal.Rest.ApplicationCommands as R
-import Discord.Internal.Types.ApplicationCommands (InteractionResponse(InteractionResponse))
+import Discord.Internal.Types.ApplicationCommands 
 import Data.Aeson
 
 -- Allows this code to be an executable. See discord-haskell.cabal
@@ -100,8 +99,8 @@ eventHandler event = case event of
                       (Just True) 
                       (Just 
                         [
-                          ApplicationCommandOptionChoice "firstopt" (SIDS "yay"), 
-                          ApplicationCommandOptionChoice "secondopt" (SIDS "nay")
+                          ApplicationCommandOptionChoice "firstopt" (SNS "yay"), 
+                          ApplicationCommandOptionChoice "secondopt" (SNS "nay")
                         ]
                       ) Nothing Nothing Nothing Nothing Nothing 
                   ]
@@ -111,14 +110,14 @@ eventHandler event = case event of
             ) 
           )
         ) >>= \rs -> trace (show rs) (return ())
-      InteractionCreate i@(Interaction {..}) ->  trace (show i) $ do
+      InteractionCreate i@Interaction {..} ->  trace (show i) $ do
         let cid = fromJust interactionChannelId
-        if not (interactionType == APPLICATION_COMMAND) then void $ restCall (R.CreateMessage cid "I don't know how to handle an interaction like that!")
+        if interactionType /= APPLICATION_COMMAND then void $ restCall (R.CreateMessage cid "I don't know how to handle an interaction like that!")
         else do
           let d = fromJust interactionData
           case interactionDataApplicationCommandType d of
             ACTCHAT_INPUT -> void $  restCall (
-                let x = (InteractionResponse 
+                let x = InteractionResponse 
                             CHANNEL_MESSAGE_WITH_SOURCE 
                             (Just 
                               (ICDM 
@@ -126,7 +125,7 @@ eventHandler event = case event of
                                   Nothing 
                                   (Just 
                                     (T.pack $ "Here's the reply! You chose: " 
-                                      ++ (show $ fromJust $ applicationCommandInteractionDataOptionValue $ head  (fromJust $ interactionDataOptions d))
+                                      ++ show (fromJust $ applicationCommandInteractionDataOptionValue $ head  (fromJust $ interactionDataOptions d))
                                     )
                                   )
                                   Nothing
@@ -136,7 +135,7 @@ eventHandler event = case event of
                                 )
                               )
                             )
-                          ) in trace (show (toJSON x)) (R.CreateInteractionResponse interactionId interactionToken x)
+                          in trace (show (toJSON x)) (R.CreateInteractionResponse interactionId interactionToken x)
                 )
             _ -> void $ restCall (R.CreateMessage cid "I got some other kind of application command!")
       -- Note that the above is not the required way of receiving and replying to interactions - this is marked as a failure
--- a/examples/ping-pong.hs
+++ b/examples/ping-pong.hs
@@ -122,7 +122,7 @@ eventHandler event = case event of
                             CHANNEL_MESSAGE_WITH_SOURCE 
                             (Just 
                               (ICDM 
-                                (InteractionCallbackDataMessages 
+                                (InteractionCallbackMessages 
                                   Nothing 
                                   (Just 
                                     (T.pack $ "Here's the reply! You chose: " 
--- a/examples/ping-pong.hs
+++ b/examples/ping-pong.hs
@@ -1,20 +1,19 @@
-{-# LANGUAGE OverloadedStrings #-}  -- allows "strings" to be Data.Text
-{-# LANGUAGE  RecordWildCards #-}
+-- allows "strings" to be Data.Text
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
 
-import Control.Monad (when, forM_, void)
+import Control.Monad (forM_, void, when)
+import Data.Aeson
+import Data.Maybe (fromJust)
 import qualified Data.Text as T
 import qualified Data.Text.IO as TIO
-
+import Debug.Trace
+import Discord
+import Discord.Internal.Types.ApplicationCommands
+import qualified Discord.Requests as R
+import Discord.Types
 import UnliftIO (liftIO)
 import UnliftIO.Concurrent
-
-import Discord
-import Discord.Types
-import qualified Discord.Requests as R
-import Debug.Trace
-import Data.Maybe (fromJust)
-import Discord.Internal.Types.ApplicationCommands 
-import Data.Aeson
 
 -- Allows this code to be an executable. See discord-haskell.cabal
 main :: IO ()
@@ -26,12 +25,15 @@ pingpongExample = do
   tok <- TIO.readFile "./examples/auth-token.secret"
 
   -- open ghci and run  [[ :info RunDiscordOpts ]] to see available fields
-  t <- runDiscord $ def { discordToken = tok
-                        , discordOnStart = startHandler
-                        , discordOnEnd = liftIO $ putStrLn "Ended"
-                        , discordOnEvent = eventHandler
-                        , discordOnLog = \s -> TIO.putStrLn s >> TIO.putStrLn ""
-                        }
+  t <-
+    runDiscord $
+      def
+        { discordToken = tok,
+          discordOnStart = startHandler,
+          discordOnEnd = liftIO $ putStrLn "Ended",
+          discordOnEvent = eventHandler,
+          discordOnLog = \s -> TIO.putStrLn s >> TIO.putStrLn ""
+        }
   TIO.putStrLn t
 
 -- If the start handler throws an exception, discord-haskell will gracefully shutdown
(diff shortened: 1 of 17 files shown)


src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -0,0 +1,396 @@
...
+
+instance ToJSON ApplicationCommandOptionType where
+  toJSON = toJSON . fromEnum
+
+instance FromJSON ApplicationCommandOptionType where
+  parseJSON = withScientific "ApplicationCommandOptionType" (return . toEnum . round)
+
+-- | Utility data type to store integers or doubles.
+data IntDouble = IDI Int | IDD Double
+  deriving (Show, Read, Eq)
```
aquarial, line 306: Is this utility necessary? A 64bit Double can represent ~52 bit integers 

https://stackoverflow.com/questions/759201/representing-integers-in-doubles

src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -0,0 +1,396 @@
...
+
+instance ToJSON IntDouble where
+  toJSON (IDI s) = toJSON s
+  toJSON (IDD i) = toJSON i
+
+instance FromJSON IntDouble where
+  parseJSON v = (IDI <$> parseJSON v) <|> (IDD <$> parseJSON v)
+
+-- | Utility data type to store strings or number types.
+data StringNumber = SNS String | SNN IntDouble
```
aquarial, line 316: Hmm. I'd prefer the ADT name prefixed, but that would be an ugly name (`StringNumberString`?) Hmm.

src/Discord/Internal/Types/Interactions.hs
```diff
@@ -0,0 +1,363 @@
...
+import Discord.Internal.Types.Prelude (ApplicationId, ChannelId, GuildId, Snowflake)
+import Discord.Internal.Types.User (User)
+
+toMaybeJSON :: (ToJSON a) => a -> Maybe Value
+toMaybeJSON = return . toJSON
+
+makeTable :: (Data t, Enum t) => t -> [(Int, t)]
+makeTable t = map (\cData -> let c = fromConstr cData in (fromEnum c, c)) (dataTypeConstrs $ dataTypeOf t)
+
+type InteractionId = Snowflake
```
aquarial, line 44: Move to internal.types.prelude

src/Discord/Internal/Types/Interactions.hs
```diff
@@ -0,0 +1,363 @@
...
+            <*> v .:? "member"
+            <*> v .:? "user"
+            <*> v .: "token"
+            <*> v .: "version"
+            <*> v .:? "message"
+      )
+
+-- | What type of interaction has a user requested? Each requires its own type
+-- of response.
+data InteractionType = PING | APPLICATION_COMMAND | MESSAGE_COMPONENT | APPLICATION_COMMAND_AUTOCOMPLETE
```
aquarial, line 106: Prefix ADT name

src/Discord/Internal/Types/Interactions.hs
```diff
@@ -0,0 +1,363 @@
...
+data ApplicationCommandInteractionDataOption = ApplicationCommandInteractionDataOption
+  { applicationCommandInteractionDataOptionName :: String,
+    applicationCommandInteractionDataOptionType :: ApplicationCommandOptionType,
+    -- | The value itself. Mutually exclusive with options.
+    applicationCommandInteractionDataOptionValue :: Maybe StringNumber,
+    -- | Only present in group subcommands and subcommands. Mutually exclusive with value.
+    applicationCommandInteractionDataOptionOptions :: Maybe [ApplicationCommandInteractionDataOption],
+    -- | Whether this is the field that the user is currently typing in.
+    applicationCommandInteractionDataOptionFocused :: Maybe Bool
+  }
```
aquarial, line 225: :+1:  for the names. Tough to put "mutually exclusive with value" in a comment. Ideally this is abstracted into an ADT to express this, but I get that this is all WIP on new docs. Not worth it now.

src/Discord/Internal/Types/Interactions.hs
```diff
@@ -0,0 +1,363 @@
...
+        | (name, Just value) <-
+            [ ("type", toMaybeJSON interactionResponseType),
+              ("data", toJSON <$> interactionResponseData)
+            ]
+      ]
+
+-- | What's the type of the response?
+data InteractionCallbackType
+  = -- | Responds to a PING.
+    PONG
```
aquarial, line 279: Prefix ADT names.

src/Discord/Internal/Types/Interactions.hs
```diff
@@ -0,0 +1,363 @@
...
+
+-- | Convenience wrapper for two separate types of callback.
+data InteractionCallbackData = ICDM InteractionCallbackMessages | ICDA InteractionCallbackAutocomplete
+  deriving (Show, Read, Eq)
+
+instance ToJSON InteractionCallbackData where
+  toJSON (ICDM icdm) = toJSON icdm
+  toJSON (ICDA icda) = toJSON icda
+
+type InteractionCallbackAutocomplete = [ApplicationCommandOptionChoice]
```
aquarial, line 315: What's the use for this?

src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -0,0 +1,396 @@
...
+    GUILD_STORE
+  | -- | A temporary sub-channel within a guild_news channel.
+    GUILD_NEWS_THREAD
+  | -- | A temporary sub-channel within a guild_text channel
+    GUILD_PUBLIC_THREAD
+  | -- | A temporary sub-channel within a GUILD_TEXT channel that is only
+    -- viewable by those invited and those with the MANAGE_THREADS permission
+    GUILD_PRIVATE_THREAD
+  | -- | A voice channel for hosting events with an audience.
+    GUILD_STAGE_VOICE
```
aquarial, line 373: Prefix ADT name

examples/ping-pong.hs
```diff
@@ -1,15 +1,19 @@
...
 import qualified Data.Text as T
 import qualified Data.Text.IO as TIO
-
-import UnliftIO (liftIO)
-import UnliftIO.Concurrent
-
+import Debug.Trace
 import Discord
-import Discord.Types
+import Discord.Internal.Types.ApplicationCommands
```
aquarial, line 12: The point if `Internal` is that the user shouldn't import it. It's a good idea to show how to use the internal stuff, but maybe in a more advanced example. pingpong has enough going on already.

aquarial (commented):
Haskell creates a function for each ADT field, so they have to be unique. Short field names 'pollute' the namespace with irrelevant options.

In discord-haskell I use a convention where I always prefix the ADT name before each field. It's long and unwieldy, but consistent, discoverable, and autocompleteable. 

`STRING` is an important identifier, and shouldn't be used for an `ApplicationCommandOptionType` variant. 

.

Ugh, annoying that `Guild Member` doesn't always attach the `User` field. This is something the `InternalEvent` + `Cache` could be really helpful for. 

.

> -- | I'm not sure what this stuff is, so you're on your own.

Hahah yeah I feel this. So much of the docs are spaghetti normalized, and spread out into "maybe this can happen" optional fields. 

.

The autoformatter looks good in these examples. Open to more in the future.

.


I want to merge this into the repo, it's not ready for master branch yet. Merge into a new github branch and I'll add a note in the master Readme? Stack can build a specific git commit as a dep, so someone who needs the functionality can try it out, but I don't want someone to accidentally use is yet.

src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -0,0 +1,396 @@
...
+
+-- | What type of application command. Represents slash commands, right clicking
+-- a user, and right clicking a message respectively.
+data ApplicationCommandType
+  = -- | Slash commands
+    ACTCHAT_INPUT
+  | -- | User commands
+    ACTUSER
+  | -- | Message commands
+    ACTMESSAGE
```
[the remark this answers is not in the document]
L0neGamer: Would you also want them un-capitalised and un-underscored?

src/Discord/Internal/Rest/ApplicationCommands.hs
```diff
@@ -0,0 +1,87 @@
...
+import Discord.Internal.Types.ApplicationCommands
+import Network.HTTP.Req as R
+
+instance Request (ApplicationCommandRequest a) where
+  jsonRequest = applicationCommandJsonRequest
+  majorRoute = applicationCommandMajorRoute
+
+type AppCom = ApplicationCommand
+
+type CreateAppCom = CreateApplicationCommand
```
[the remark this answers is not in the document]
L0neGamer: They can be removed, honestly

src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -0,0 +1,396 @@
...
+    cApplicationCommandDescription :: String,
+    -- | What options the application (max length 25). Has to be `Nothing` for
+    -- non-slash  commands.
+    cApplicationCommandOptions :: Maybe [ApplicationCommandOption],
+    -- | Whether the command is enabled by default when the application is added
+    -- to a guild. Defaults to true if not present
+    cApplicationCommandDefaultPermission :: Maybe Bool,
+    -- | What the type of the command is. If `Nothing`, defaults to slash
+    -- commands.
+    cApplicationCommandType :: Maybe ApplicationCommandType
```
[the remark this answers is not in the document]
L0neGamer: My concern is that some of the ADT names are _incredibly_ long already

src/Discord/Internal/Types/ApplicationCommands.hs
L0neGamer: I guess the question is do we want to spit out 64bit doubles all the time or allow some things to be integers or doubles?

src/Discord/Internal/Types/Interactions.hs
L0neGamer: It makes it clearer what the option is in terms of types; it can be removed if wanted.

src/Discord/Internal/Types/Interactions.hs
L0neGamer: My plan is for _all_ of this stuff to be made "Internal", and then create ADTs that translate into them, so the internals are replicas of the input JSON and the ones that we advertise to users are usable in haskell.

src/Discord/Internal/Types/ApplicationCommands.hs
L0neGamer: I can try to come up with an alternative name ig

examples/ping-pong.hs
L0neGamer: Ah, sorry, my code complete must of automatically grabbed from all the way in the internals.

L0neGamer:
Closing so it can be reopened on a new branch (name `dev` maybe?).
I agree with the comments here, but there are definitely aspects that it would be useful to have guidance on (like is it a good idea to have 10-15 character prefixes for certain things?)

src/Discord/Internal/Types/ApplicationCommands.hs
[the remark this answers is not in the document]
aquarial: Yeah. It's a convention I'm using for this product, but I'm not sure if it's any standard.

src/Discord/Internal/Types/ApplicationCommands.hs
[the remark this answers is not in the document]
aquarial: Hmm. `CreateApplicationCommandDefaultPermission` ?  It's long but explicit, and shouldn't be used that often anyway. 

(I come from a Java background so maybe I'm super biased here lm).

src/Discord/Internal/Types/ApplicationCommands.hs
aquarial: Good point. Should probably allow `int`, `double`, `string` different options. This is find for the backend, but maybe the user frontend could abstract using a nicer ADT? 

`ApplicationCommandOption` has some redundancies, like what happens if the type is set to `string`, but `int` choices are given? And `optionMinVal` for string. 

Fine for the backend, (match the json structure), but ideally the frontend is nicer eventually.

src/Discord/Internal/Types/Interactions.hs
aquarial: That's a good idea

src/Discord/Internal/Types/Interactions.hs
aquarial: Hmm. Okay to leave for now.

aquarial:
:+1: 

Want to say thanks for putting this together, along with SocialAnthrax. It's a lot of code and details to get into place, a lot of work.

The 15char prefix question is tough. I lean towards long names because there are a lot of nested ADTs and I want to make it explicit which field goes into which ADT. It's not usual Haskell style though, so I'm also not fully sure what to do.

src/Discord/Internal/Types/ApplicationCommands.hs
L0neGamer: I've come up with the solution (which Aeson already had funnily enough)

We just use the Scientific type, and then people can convert out of that if they wish.

When we get around to making a nice front end for all this stuff, we can lock in the choices, but I think this is the best approach for now - honestly "not caring" about the type being given to the JSON, since JSON doesn't differentiate between number and integer anyway

src/Discord/Internal/Types/Interactions.hs
L0neGamer: Just the type, or the `makeTable` too?

aquarial: Oh true, just the type, since `makeTable` is a utility only used in this file.

PR #86, opened 2021-12-18
Tidying and neatening

L0neGamer:
Fixed a bunch of minor things, and added CI (which currently fails) to the repo.

The CI checks that the files are all formatted using ormolu (which has been mentioned in the README).

https://github.com/WarwickTabletop/tablebot is what I'm basing the CI off of, and may be useful from a documentation standard stand point.


aquarial:
neat!

PR #88, opened 2021-12-19
Application commands

L0neGamer:
Another day, another PR.

Things that still need to be worked on - type safe front end for easy use, add components so that the API is more complete.

### DIFF
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -1,7 +1,7 @@
 cabal-version:       2.0
 name:                discord-haskell
 -- library version is also noted at src/Discord/Rest/Prelude.hs
-version:             1.9.1
+version:             1.9.2
 description:         Functions and data types to write discord bots.
                      Official discord docs <https://discord.com/developers/docs/reference>.
                      .
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -1,7 +1,7 @@
 cabal-version:       2.0
 name:                discord-haskell
 -- library version is also noted at src/Discord/Rest/Prelude.hs
-version:             1.8.9
+version:             1.9.0
 description:         Functions and data types to write discord bots.
                      Official discord docs <https://discord.com/developers/docs/reference>.
                      .
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -1,7 +1,7 @@
 cabal-version:       2.0
 name:                discord-haskell
 -- library version is also noted at src/Discord/Rest/Prelude.hs
-version:             1.9.0
+version:             1.9.1
 description:         Functions and data types to write discord bots.
                      Official discord docs <https://discord.com/developers/docs/reference>.
                      .
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -34,7 +34,6 @@ executable ping-pong
                      , text
                      , unliftio
                      , discord-haskell
-                     , aeson
 
 library
   ghc-options:         -Wall
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -34,6 +34,7 @@ executable ping-pong
                      , text
                      , unliftio
                      , discord-haskell
+                     , aeson
 
 library
   ghc-options:         -Wall
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -94,3 +94,4 @@ library
                      , wuss
                      , mtl
                      , unliftio
+                     , scientific
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -60,6 +60,7 @@ library
                      , Discord.Internal.Rest.Voice
                      , Discord.Internal.Rest.Webhook
                      , Discord.Internal.Rest.ApplicationCommands
+                     , Discord.Internal.Rest.Interactions
                      , Discord.Internal.Types
                      , Discord.Internal.Types.Prelude
                      , Discord.Internal.Types.Channel
@@ -69,6 +70,7 @@ library
                      , Discord.Internal.Types.User
                      , Discord.Internal.Types.Embed
                      , Discord.Internal.Types.ApplicationCommands
+                     , Discord.Internal.Types.Interactions
   build-depends:
                        base >= 4 && <5
                      , aeson
--- a/examples/ping-pong.hs
+++ b/examples/ping-pong.hs
@@ -13,8 +13,7 @@ import Discord.Types
 import qualified Discord.Requests as R
 import Debug.Trace
 import Data.Maybe (fromJust)
-import qualified Discord.Internal.Rest.ApplicationCommands as R
-import Discord.Internal.Types.ApplicationCommands (InteractionResponse(InteractionResponse))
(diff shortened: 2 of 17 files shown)


discord-haskell.cabal
```diff
@@ -34,6 +34,7 @@ executable ping-pong
                      , text
                      , unliftio
                      , discord-haskell
+                     , aeson
```
aquarial, line 37: does the example ping-pong need to import aeson?

examples/ping-pong.hs
```diff
@@ -38,50 +43,158 @@ startHandler :: DiscordHandler ()
...
+      ( \channel ->
+          restCall $
+            R.CreateMessage
+              (channelId channel)
+              "Hello! I will reply to pings with pongs"
+      )
+
+-- | An example of the example slash command using the defaults
+exampleSlashCommand' :: CreateApplicationCommand
+exampleSlashCommand' =
```
aquarial, line 75: Document that this is unused, and is equivalent (?) to the other exampleCommand

examples/ping-pong.hs
```diff
@@ -38,50 +43,158 @@ startHandler :: DiscordHandler ()
...
+  Ready _ _ _ _ _ _ pa@(PartialApplication i _) ->
+    trace
+      (show pa)
+      ( restCall
+          ( R.CreateGuildApplicationCommand i serverid exampleSlashCommand
+          )
+      )
+      >>= \rs -> trace (show rs) (return ())
+  InteractionCreate i@Interaction {..} -> trace (show i) $ do
+    let cid = fromJust interactionChannelId
```
aquarial, line 181: Is `fromJust` safe here? A crash in an event handler is ignored so in general yes, but what's the particular thought here?

src/Discord/Internal/Types/Interactions.hs
```diff
@@ -0,0 +1,365 @@
...
+  toEnum a = fromJust $ lookup a table
+    where
+      table = makeTable InteractionCallbackTypePong
+
+instance ToJSON InteractionCallbackType where
+  toJSON = toJSON . fromEnum
+
+-- | Convenience wrapper for two separate types of callback.
+data InteractionCallbackData = ICDM InteractionCallbackMessages | ICDA InteractionCallbackAutocomplete
+  deriving (Show, Read, Eq)
```
aquarial, line 311: `ICDM` and `ICDA` are weird. 

```haskell
data InteractionCallbackData =  InteractionCallbackDataMessage InteractionCallbackMessages 
                             | InteractionCallbackDataAutocomplete InteractionCallbackAutocomplete
```

Long unwieldy, but explicit. It's uncommon enough that I prefer to follow the ugly rule than create a short exception

src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -0,0 +1,382 @@
...
+      ]
+
+-- | Data type to be used when editing application commands. The specification
+-- is below. See `CreateApplicationCommand` for an explanation for the
+-- parameters.
+--
+-- https://discord.com/developers/docs/interactions/application-commands#edit-global-application-command
+data EditApplicationCommand = EditApplicationCommand
+  { editApplicationCommandName :: Maybe String,
+    editApplicationCommandDescription :: Maybe String,
```
aquarial, line 117: Use Text instead of String. (I prefer to use text for types as much as possible)

src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -0,0 +1,382 @@
...
+  parseJSON =
+    withObject
+      "ApplicationCommand"
+      ( \v ->
+          ApplicationCommand
+            <$> v .: "id"
+            <*> v .:? "type"
+            <*> v .: "application_id"
+            <*> v .:? "guild_id"
+            <*> v .: "name"
```
aquarial, line 176: Oh neat, this is a reasonable way to format https://github.com/aquarial/discord-haskell/issues/87. The `.:` v `.:?` formatting is still a minor complaint, but my main gripe is fixed here

src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -0,0 +1,382 @@
...
+-- | This is the structure that designates different options for slash commands.
+--
+-- https://discord.com/developers/docs/interactions/application-commands#application-command-object-application-command-option-structure
+data ApplicationCommandOption = ApplicationCommandOption
+  { -- | What the type of this option is.
+    applicationCommandOptionType :: ApplicationCommandOptionType,
+    -- | The name of the option . 1-32 characters
+    applicationCommandOptionName :: String,
+    -- | 1-100 characters
+    applicationCommandOptionDescription :: String,
```
aquarial, line 192: String --> Text (and elsewhere)

src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -0,0 +1,382 @@
...
+
+instance ToJSON ApplicationCommandOptionType where
+  toJSON = toJSON . fromEnum
+
+instance FromJSON ApplicationCommandOptionType where
+  parseJSON = withScientific "ApplicationCommandOptionType" (return . toEnum . round)
+
+-- | Utility data type to store strings or number types.
+data StringNumberValue = StringNumberValueString String | StringNumberValueNumber Scientific
+  deriving (Show, Read, Eq)
```
aquarial, line 303: Looks nicer than Int/Double situation! 

String->Text for values. I think it's ok to do `StringNumberValueString T.Text` because text is a string type. Open to `TextNumberValueText Text` or `Either Text Scientific`

aquarial (commented):
Message, User, and GuildMember ADT are very impressive!

Overall nice stuff. 

Noticed that pull https://github.com/aquarial/discord-haskell/pull/90 is downstream of this, will review that as well

examples/ping-pong.hs
L0neGamer: The thought was to give a quick and dirty demonstration of how interactions could work, although I accept that this is unsafe (as is `head`).

I can come up with safer alternatives, but I didn't want to have to mess with the monads.

src/Discord/Internal/Types/ApplicationCommands.hs
L0neGamer: Other formatters may even fix this, but I've been focussing more on implementation tbh.

src/Discord/Internal/Types/ApplicationCommands.hs
L0neGamer: I think the name should still be string to reflect the JSON descriptions of such things

examples/ping-pong.hs
aquarial: Yeah it's not necessary, but I prefer to have good practices in examples. Don't have to change it now

src/Discord/Internal/Types/ApplicationCommands.hs
aquarial: :+1: 

Implementation is looking good

aquarial:
Neat!

PR #90, opened 2021-12-19
Components

L0neGamer:
Adding components support; had to move a fair amount of stuff around to prevent cyclical dependencies.

SendMessageDetailed needs to be expanded as a result, and I still need to create the lib-user-facing niceties to prevent invalid states.


L0neGamer:
No need to review just yet - wait until #88 is done.

L0neGamer:
This update... may break things.

I've changed some of the pertinent stuff in the code for v8 but I can't be sure to have caught everything.

A complete go through of the code needs to be done, and we likely need to annotate data structures and functions with what version they were last correct on.

L0neGamer:
I'm gonna close this so I can finish off making things much nicer for lib users

PR #98, opened 2022-02-06
Should fix the argument type issue

### LINKED ISSUE #97: ApplicationCommandOptionValueString sets up integer argument
When I use `ApplicationCommandOptionValueString` to set up a String option to my application command, discord says it requires an integer argument

```hs
 ApplicationCommandOptionValueString
        "group"
        "Group to get the planning for"
        (Just True)
        Nothing
        Nothing
 ```
image

I think the problem is here: https://github.com/aquarial/discord-haskell/blob/b14ee1843ec1be5ba5cbfac6e1c2302d9284d50c/src/Discord/Internal/Types/ApplicationCommands.hs#L153 as it uses `ApplicationCommandOptionTypeInteger` in the wrapper, though I may be wrong

Annwan:
Fixes #97, hopefully


Annwan:
Tested on my bot and it seems to fix the problem

aquarial:
:+1: 

I'm wrangling through https://github.com/aquarial/discord-haskell/pull/96, which moves the code this pr changes somewhere else, will keep this bug in mind when looking at it.

Thanks for looking into the error! Makes fixing things much easier for me

Annwan:
Oh nice, I did look at that pull request but it moved a lot of stuff so I had a hard time keeping track of stuff
I'll use my fork as a temporary fix in the mean time

Feel free to close this whenever L0neGamer's PR is merged

L0neGamer:
I believe this is the equivalent? 
https://github.com/L0neGamer/discord-haskell/blob/interactions-neatening/src/Discord/Internal/Types/ApplicationCommands.hs#L331-L338

Annwan:
> I believe this is the equivalent? https://github.com/L0neGamer/discord-haskell/blob/interactions-neatening/src/Discord/Internal/Types/ApplicationCommands.hs#L331-L338

Yeah looks good to me
Can't wait to rewrite all this part of the bot when the update is merged, because that looks like a way more sane than the current way

L0neGamer:
My original thoughts were "I will be as close to the API as possible", but that makes the bot nigh unusable, and doesn't take advantage of haskell typing. So I then introduced a front end version of a couple types, but to make things "simpler" I made a way to convert them. the latest PR goes "screw it, we don't need the internal version, just parse the good, type safe ones".

basically discord api is really general which does not lend itself to type safety

PR #99, opened 2022-02-07
Include embed part in `CreateMessageDetailed`

matobet:
After migrating from:
```haskell
DR.CreateMessageEmbed channelId body $ def { DT.createEmbedImage = Just ... }
```
to
```haskell
DR.CreateMessageDetailed
              channelId
              def
                { DT.messageDetailedContent   = body
                , DT.messageDetailedEmbed     = Just $ def { DT.createEmbedImage = Just ... }
                }
```
I noticed pictures posted by my bot broke (became small black-ish squares).

After some investigation it seems that `CreateMessageDetailed` is not correctly passing the embed part as a part of the multi-part request (only the passed file). This PR fixes this.


aquarial:
:+1: 

Thanks

PR #104, opened 2022-02-22
Modals

L0neGamer:
added the new modal interaction, as well as the required components. Some of the parsing is a bit messy, but that's because I wanted to limit what could go into a modal (as what can go into a modal is limited already).

contains #103 as well, so that should be handled first.


aquarial:
Wow! I really appreciate you going through all this documentation and getting all the types to line up.

Adding (Ord+Read+Show+Eq) instances to all the interactions is a nice touch as well.

What else changed between v8 and v10? (really v6 was the last time I read through things). Is it just adding new features?

L0neGamer:
V8: https://discord.com/developers/docs/change-log#api-and-gateway-v8. 
V9: https://discord.com/developers/docs/change-log#api-v9. 
V10 has a couple small changes from what I remember.  Biggest changes I've noticed are requiring certain intents in V10 which may catch some people out (I know it caught me out).

aquarial:
:+1:

Issue #106, opened 2022-03-02
EditMessage is Outdated


XanderDJ:
With the new MessageDetailedOpts you can add stuff like actionrowcomponents following the api of discord. But this functionality isn't present in the EditMessage Channel Request. It's not possible to speciffy what you want to do for components in that end point. While this should be possible discord wise I think. Current endpoint
![image](https://user-images.githubusercontent.com/26298748/156455399-663e5993-b273-46cc-93bb-bd429b03bc38.png)

CreateEmbed specifies no fields for components like the MessageDetailedOpts does:
![image](https://user-images.githubusercontent.com/26298748/156455572-64f9e695-3e87-4245-bf5d-b3c629768f32.png)

When handling the code itself it's using a field in JSON that will be deprecated in the future namely _embed_ instead of embeds.

Edit endpoint according to discord docs:
![image](https://user-images.githubusercontent.com/26298748/156456096-baeff4b2-b7c0-4f33-bef5-9f5e2bb9e01f.png)
It has the components attributes.

aquarial:
Should be fixed by https://github.com/aquarial/discord-haskell/pull/107, pushed int 1.12.2

PR #107, opened 2022-03-03
updating some message channel stuff

L0neGamer:
Should solve #106, and clean up just a tad.

### DIFF
--- a/examples/ping-pong.hs
+++ b/examples/ping-pong.hs
@@ -71,7 +71,8 @@ eventHandler event = case event of
         threadDelay (2 * 10 ^ (6 :: Int))
 
         -- A very simple message.
-        void $ restCall (R.CreateMessage (messageChannelId m) "Pong!")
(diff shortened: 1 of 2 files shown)


aquarial:
Neat work! 

Filled in some of the fields too which is good

yutotakano:
If you weren't considering it already @aquarial, it might be worth adding such removals (`CreateMessageUploadFile`) and changes in public-facing type signature to the Changelog as well! `discord-haskell`'s Changelog tends to be quite vague in its changes, but in my opinion it's often quite useful to have a bit more detail because I can see which parts are removed/deprecated/added when upgrading the library. (The alternative would be to compile and pray, or to dive into each commit themselves)

aquarial:
True! 

I forgot to update the changelog for this one. 

The change log links to the pull-request, so it's a bit inconvenient to require a pullrequest to link to itself. Could also change the format of the changelog to not have links. (or link to the issue).

PR #110, opened 2022-03-18
Emoji/sticker update

L0neGamer:
Adding sticker components and moving emoji stuff around.

untested due to inability to create and manipulate stickers in my test server.


aquarial:
:+1: neat


There's no public testing server with stickers or anything, so this will do for now: https://discord.com/channels/81384788765712384/381870553235193857/954490421386563684

PR #113, opened 2022-04-07
Monad discord

L0neGamer:
Creates a monad discord type class, and defines it for DiscordHandler (which should still work just drop-in), IdentityT, and ReaderT. Further instances like StateT aren't possible due to the MonadUnliftIO requirement in runDiscordM.

This should allow more complex monad stacks that don't need manual lifting.

To use a `MonadDiscord m`, use `runDiscordM` with a function that "unlifts" the `m` to a `DiscordHandler` - so in the case of the defined instances, means supplying a context.

Unfortunately since this effectively moves all the processing code the history of `Discord.hs` won't be around - ~this might be fixable with git trickery?~ tried some git trickery but it didn't work, oh well.


aquarial:
This moves code from src/Discord.hs into src/Discord/Monad to hide types right? So someone who wants to use runDiscordM could import it, but the direct usage wouldn't have to? What's the use-case there? I'm trying to figure out what this code change *means*, but I don't get monad transformers. 

What's the intuition with `MonadDiscord m` stuff, when would it show up in an end-user type signature?

Is there a backwards compatibility problem where making the current src/discord.hs as generic as the new src/discord/monad would cause type errors? This PR doesn't change any of the examples,  which is a good thing in other ways, but also I don't know enough about MTL to figure out why this is changing things.

L0neGamer:
The use case is being able to stack ReaderT (and IdentityT but that's useless) classes, and being able to use them and their context throughout the process without having to lift yourself, or to do weird state.

For example, say you wanted to use a data structure as a context throughout your program. Using MonadDiscord, you could just define `type MyHandler = ReaderT MyData DiscordHandler`, and use all the other functions of discord haskell within this new context, while still being able to access your provided MyData.

Another advantage is as follows. Say that `type MyData  = MVar MoreData`. Due to how `runDiscordM` (and helpers) work, the context provided by MyData can then be modified during `discordOnStart`.

Defining this type class also allows others to create their own implementations of things if they want.

This move to MonadDiscord could clean up how people use this library.

aquarial:
I don't want to merge this. I've been thinking about this and I prefer https://github.com/aquarial/discord-haskell/pull/112/ handle IO instead of DiscordHandler MTL.

The best compromise I see is to split into two separate libraries: One uses IO handles, and one uses the MTL interface. One can depend on the other, sharing almost all the code. Can discuss what the libraries are if you're interested (`discord-haskell-mtl` or `discord-haskell-io`?)

-------------------------------

It's great code but I'm conflicted.

Exposing the MTL lego bricks is a natural evolution, and adding the abstraction in a background module (with backwards compatibility) is nice. It's neat and clever, all the types check out.

Overall it fits with one idea of what "good" haskell code looks like: bring things into the type level.

However I don't like it at all. I see another layer of type-indirection laying a muddy ball of Monads and lifts ontop of OS concurrency primitives like MVar/Chan. I've read a lot about monad transformers and I'm not interested, but I also know many other people like them and think they're the best style of haskell.

Two blogposts about haskell design

- https://www.fpcomplete.com/blog/2017/06/readert-design-pattern/
  - suggests using `ReaderT Config IO` for apps for simplicity

- https://jaspervdj.be/posts/2018-03-08-handle-pattern.html
  - use a `Handle` for simple `IO ()`


-----------------------------------


The code is great (it's a worthy goal and other people agree with it) but I don't like modern haskell style, and I don't want to have MTL in the library.

Two people liked <https://github.com/aquarial/discord-haskell/pull/112#issuecomment-1074871635> about not merging https://github.com/aquarial/discord-haskell/pull/112

L0neGamer:
My view on a split is that doing so would result in more work trying to keep those libraries going. If we wanted to have both MTL stuff and IO-handle stuff, I'd honestly prefer modules like discord.monad and discord.io that instead of replacing the standard callings, just use them, and if people would prefer to use them then the import that module instead.

For example, the discordmonad restcall' is basically just the normal restcall. If that was imported in the qualified way, then having it defined in-module is not required.

This would mean that there would be conflicting definitions for some of the base types and functions, but users should only be using one of each of the modules (`Discord` to continue as normal, `Discord.Monad` to use the MTL stuff, and `Discord.IO` to use the IO functions).

Reading the articles you posted. The first ends off with
> So the take-away here is: if you can generalize your functions to mtl-style Monad constraints, do it, you'll regain a lot of the benfits you'd have with purity.
Which seems in favour of the MTL approach. Importantly, the article isn't necessarily against having transformers for things.

It also gave me the idea just now of making the base MonadDiscord class pure, and then one can make testing versions that don't require IO... I'll work on this at some point.

I read some of the second article, and while there may be benefits to the handle-as-argument pattern, the ReaderT pattern handles (heh) any need for passing context around. What's more, it means users of the library never have to touch this important and, ultimately, useless to them data structure.

aquarial:
> while there may be benefits to the handle-as-argument pattern, the ReaderT pattern handles (heh) any need for passing context around. What's more, it means users of the library never have to touch this important and, ultimately, useless to them data structure

Hm I can appreciate the `ReaderT Handle IO` because it passes the handle around as implicit context. Does seem to be "good" Haskell style, and a lot of people are really into it. But I don't like it. 

Maybe because I write code in other languages a lot. Explicitly passing the arguments into `readCache :: Handle -> IO Cache` is the same as C style `cache * readCache(Handle * h)`, or even the java style `cache = handle.readCache()`. 

The question is how to view the function of `readCache`. Does it require an implicit Discord context that can come from anywhere (MTL monad stack), or is it running operations on a specific handle (handle). 

File IO uses the handle UI. To read you call `openFile :: FilePath -> IOMode -> IO Handle` and then use handle functions like `hGetContents :: Handle -> IO Text` that take a specific handle and run IO. It makes sense to have multiple files open and want to read the contents of one specific file using it's handle so the operation needs to take the handle as an argument. 

That's how I think about the DiscordHandle here. I want to pass the handle in because I think of `readCache` as running on the handle. But if you think of `readCache :: MonadDiscord m => m Cache` as an action to run to get the cache that just requires the context of discord to be available, then the MTL makes more sense.

------------------------------------------------

Theorycrafting here for  second:

If you make the type-signature something general then readCache could be compatibile with monad stacks, while becoming concrete if someone sticks with DiscordHandler.

```haskell
readCache :: (MonadConc m, HasDiscordHandle m) => m Cache
readCache = ...


{- TODO: might be a way to define HasDiscordHandle and DiscordHandler so that the types compose here -}
class HasDiscordHandle m where
  getHandle = 

type DiscordHandler a = (MonadConc m, HasDiscordHandle m) => ReaderT Discord IO

someAction :: DiscordHandler ()
someAction = do cache <- readCache
                blaBlah
```

I don't really understand MTL enough to generate these kind of types, but it feels like there's a solution that could work.

L0neGamer:
I would describe readCache more as an action to get the cache from the context we're in. I think the handle-io way is more useful if you're working with a variety of handles in a general context. For us however, when performing actions related to working with discord we almost always want to be in a place with access to the discord functions we've created, and we only use the same handler. 

Your suggested changes to readCache is broadly what I'm trying to do. Thinking to the future for the second, there's no real reason that we need to enforce MonadConc on readCache, let alone HasDiscordHandle; if we just use MonadDiscord as I've described, someone could define their own instance of MonadDiscord with the only constraint being it has to be monadic. Think in a test environment: you could (and should) be able to simulate the generic MonadDiscord without ever needing concurrency. 

I'm currently working on the above, however I think I'm gonna run into the issue that eventually I'm gonna need functions in other files to be generalised, resulting in either a) duplicating code (which we definitely don't want to do, at least any more than we already have) or b) making this new monadic form of things the basis of the library as a whole. When I'm free I'll push what I've already done, and then I think I'll make a branch `monaddiscord-full` where I create that full conversion to freedom from IO.

L0neGamer:
The library has moved on a little bit since this PR; it's something I would still want but it's definitely something all maintainers should discuss.

PR #117, opened 2022-05-11
Simplifying

L0neGamer:
Big commit that basically just shortens a bunch of names, hopefully making the library nicer for people to use.

based off of https://github.com/aquarial/discord-haskell/pull/116, so check that one out first.


aquarial:
Minor breaking channges here (shortening all the interaction command names). Very much worth it.

PR #120, opened 2022-05-17
Add `default_member_permissions` field to CreateApplicationCommand

natsukagami:
See https://discord.com/developers/docs/interactions/application-commands#create-global-application-command
and https://discord.com/developers/docs/interactions/application-commands#create-guild-application-command.
The field is optional but was missing.


aquarial:
Nice addition!

Good to add documentation as the fields come in as well.

PR #121, opened 2022-05-22
Replace OverwriteId type with Either RoleId UserId

yutotakano:
## Summary

This pull request replaces `OverwriteId` with `Either RoleId UserId`, and removes the type flag from surrounding ADTs.

## API Changes

- `OverwriteId` is removed.
- Argument types for `EditChannelPermissions` and `DeleteChannelPermission` have changed to use `Either`.
- `ChannelPermissionsOptsType` is removed since `Either` is enough to replace this.
- The `Overwrite` ADT no longer has a `overwriteType` flag.
- The `ChannelPermissionsOpts` ADT no longer has a `channelPermissionsOptsType` flag.

## Motivation

I was trying to invoke a request to `EditChannelPermissions`. This requires an `OverwriteId` as one of its arguments, which is either a role or user id (specified by a flag inside the ADT). The official docs for this are [here](https://discord.com/developers/docs/resources/channel#edit-channel-permissions).

Before the recent ID overhaul, I could pass either UserId or RoleId with no problem since it was all the same type internally. However since the types are opaque now, I have to unwrap the Id and wrap it once again as an OverwriteId to use it. This is quite cumbersome. 

The current code is also not making full use of the types Haskell offers, since the `Either` type can solve the above problem, and actually even more by getting rid of the flag parameters!

### DIFF
--- a/src/Discord/Internal/Rest/Channel.hs
+++ b/src/Discord/Internal/Rest/Channel.hs
@@ -469,8 +469,8 @@ channelJsonRequest c = case c of
   (CreateChannelInvite chan patch) ->
       Post (channels // chan /: "invites") (pure (R.ReqBodyJson patch)) mempty
 
-  (DeleteChannelPermission chan perm) ->
-      Delete (channels // chan /: "permissions" // perm) mempty
+  (DeleteChannelPermission chan overwriteId) ->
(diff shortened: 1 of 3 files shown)


aquarial:
Very neat PR and explanation. Trimming the code types for more type safety because the Id system got more safe. 

:+1:

Issue #122, opened 2022-05-23
Requests missing in Webhook Rest API


Annwan:
In addition to the "Stack compatible webhook and GitHub compatible webhooks mentioned in #4. there is also
- [Edit Webhook Message](https://discord.com/developers/docs/resources/webhook#edit-webhook-message)
- [Delete Webhook Message](https://discord.com/developers/docs/resources/webhook#edit-webhook-message)

L0neGamer:
As discussed on discord, we're not going to support the slack and github webhooks.

PR #123, opened 2022-05-23
Webhook changes

L0neGamer:
Should work on #122.

### DIFF
--- a/src/Discord/Internal/Rest/Guild.hs
+++ b/src/Discord/Internal/Rest/Guild.hs
@@ -20,7 +20,7 @@ module Discord.Internal.Rest.Guild
 
 
 import Data.Aeson
-import Network.HTTP.Req ((/:))
+import Network.HTTP.Req ((/:), (/~))
 import qualified Network.HTTP.Req as R
 import qualified Data.Text as T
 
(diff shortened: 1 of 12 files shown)


src/Discord/Internal/Rest/Interactions.hs
```diff
@@ -13,7 +13,6 @@ import Discord.Internal.Types.Interactions
 import Network.HTTP.Client.MultipartFormData (PartM, partBS)
 import Network.HTTP.Req ((/:))
```
yutotakano, line 14: Maybe it's easier to add `(/~)` in the import list here, and replace the uses of `R./~`? The qualified operator isn't the prettiest syntax.

src/Discord/Internal/Rest/Webhook.hs
```diff
@@ -46,16 +45,19 @@ data WebhookRequest a where
   CreateWebhook :: ChannelId -> CreateWebhookOpts -> WebhookRequest Webhook
   GetChannelWebhooks :: ChannelId -> WebhookRequest [Webhook]
   GetGuildWebhooks :: GuildId -> WebhookRequest [Webhook]
-  GetWebhook :: WebhookId -> WebhookRequest Webhook
-  GetWebhookWithToken :: WebhookId -> T.Text -> WebhookRequest Webhook
-  ModifyWebhook :: WebhookId -> ModifyWebhookOpts
+  -- GetWebhook :: WebhookId -> WebhookRequest Webhook
+  GetWebhook :: WebhookId -> Maybe WebhookToken -> WebhookRequest Webhook
+  ModifyWebhook :: WebhookId -> Maybe WebhookToken -> ModifyWebhookOpts
```
aquarial, line 50: Nice to add these together, `Maybe` fits well.

From <https://discord.com/developers/docs/resources/webhook#webhook-resource> the advantage of the token is that it doesn't require logging into a bot? But the library requires logging in to send anything at all.

src/Discord/Internal/Rest/Guild.hs
```diff
@@ -20,7 +20,7 @@ module Discord.Internal.Rest.Guild
 
 
 import Data.Aeson
-import Network.HTTP.Req ((/:))
+import Network.HTTP.Req ((/:), (/~))
```
aquarial, line 23: Nice library function find!

aquarial (approved):
Looks good, Some minor merge conflicts with https://github.com/aquarial/discord-haskell/pull/119 and ready to merge

src/Discord/Internal/Rest/Webhook.hs
L0neGamer: I had a thought or two on that as well. the only place the auth is needed is in the restcallthread thing. as in the minimal.hs example you put up on discord a while ago, could we not expand support for webhook only messagin stuff by creating a separate one of those for noauth requests. means that most requests will just bounce, but gives people the option.

aquarial: I think using discord-haskell soley for an authenticated webhook http interface is an edge case. Creating a bot-application is pretty simple and unlocks everything else in the repo. If really desired minimal.hs is up as https://github.com/discord-haskell/discord-haskell/blob/master/examples/rest-without-gateway.hs

L0neGamer: I'm happy to remove the token cases if we think that they shouldn't be used.

aquarial: Yeah remove the webhook token from Get/Modify/Delete. (sending the token with ModifyWebhook [prevents changing the channelID](https://discord.com/developers/docs/resources/webhook#modify-webhook-with-token) the webhook runs in). 
Execute seems to require it though. 

Change that, and then you should have access to merge the PR

PR #124, opened 2022-05-25
Replace JuicyPixels dependency with a mimetype check

yutotakano:
## Summary

Warning: Sorry for the long post!

This is a slightly drastic pull request in that it removes JuicyPixels as a dependency completely, replacing the image decoding process with mime-type checking. This is less accurate but good enough as a heuristic, and is used by other libraries like discord.py. The removal of JuicyPixels is thought to reduce compilation time (unchecked). 

The PR also fixes the interface for these image-related endpoints, offering a consistent UI. I'd love some opinions on this!

### Current State

`discord-haskell` currently has three places where image processing happens:
- **When creating a new emoji:** The current code attempts to decode the entire image to check if it's of valid dimensions and format, then encode it as base64. This is through the `parseEmojiImage` smart constructor for `EmojiImageParsed`.
- **When creating a new sticker:** The current code does not do any decoding, and instead has a public ADT to construct for PNG, APNG, and LOTTIE filetypes. No validation occurs.
- **When modifying the current user's avatar:** The current code decodes the entire image, and converts it unconditionally to png before encoding it as base64. This is through the `parseCurrentUserAvatar` smart constructor for `CurrentUserAvatar`.

There are several problems with this:
- They are not consistent. All three should be fairly similar in logic, but the sticker processing has a completely different API.
- Implementation detail, but the sticker mime types for base64 are incorrect: APNG is image/png (not image/apng), and LOTTIE is a regular application/json (not image/lottie)
- Avatars don't need to be converted to PNG unconditionally, as Discord accepts other image types as well.
- Compiling the whole of JuicyPixels just for image decoding is quite wasteful

### The PR

This PR adds a `getMimeType :: B.ByteString -> Maybe T.Text` function in the Prelude, which uses the first few magic bytes to figure out if a file is PNG, JPEG, WEBP, or GIF.

Then, the PR makes the three image processing APIs consistent, using the smart constructor approach already used with emojis/avatars. These will check for file size constraints where specified by Discord, and for mime types, but **will not** decode the image. Invalid image dimensions or corrupt images will not be detected, and will be up to the user to check. This is documented in the Haddock comments.

This method is inspired heavily by [discord.py's method](https://github.com/Rapptz/discord.py/blob/69b4d9a4fae44bef5414c5846f6e0c4712c1fda5/discord/utils.py#L569) of checking images.

## API Changes

- The `StickerData` ADT is removed. It is replaced by a hidden ADT and the `parseStickerImage` smart constructor.
- `parseCurrentUserAvatar` is renamed to `parseAvatarImage` so that all image-related smart constructors are named as `parse***Image`.
- Image validation and dimension checking is now offloaded to the library user. `discord-haskell` will only perform superficial mime type checking.

### DIFF
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -100,7 +100,6 @@ library
                      , iso8601-time
                      , MonadRandom
                      , req
-                     , JuicyPixels
                      , safe-exceptions
                      , text
                      , time
--- a/src/Discord/Internal/Rest/Emoji.hs
+++ b/src/Discord/Internal/Rest/Emoji.hs
@@ -37,7 +37,7 @@ data EmojiRequest a where
   -- | Emoji object for the given guild and emoji ID
   GetGuildEmoji :: GuildId -> EmojiId -> EmojiRequest Emoji
   -- | Create a new guild emoji (static&animated). Requires MANAGE_EMOJIS permission.
-  CreateGuildEmoji :: GuildId -> T.Text -> EmojiImageParsed -> EmojiRequest Emoji
+  CreateGuildEmoji :: GuildId -> T.Text -> Base64Image Emoji -> EmojiRequest Emoji
   -- | Requires MANAGE_EMOJIS permission
   ModifyGuildEmoji :: GuildId -> EmojiId -> ModifyGuildEmojiOpts -> EmojiRequest Emoji
   -- | Requires MANAGE_EMOJIS permission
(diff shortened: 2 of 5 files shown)


yutotakano:
Temporarily putting the PR on hold while I evaluate a phantom-types based approach that @L0neGamer suggested to me.

L0neGamer:
I like this PR; I have a minor suggestion on the types, which is you can alias the `Base64 a` types  to specific named ones if you want for clairty, but honestly it reads completely fine to me!

I guess one question is how feasible is a basic check of the given image data? I think it is completely fine to not give any checks,  but was wondering whether it was something considered.

yutotakano:
> I like this PR; I have a minor suggestion on the types, which is you can alias the Base64 a types to specific named ones if you want for clairty, but honestly it reads completely fine to me!

Thanks! I think I agree, there's enough semantic information that "Base64Image Emoji" makes sense. We're changing the exported names anyway so I don't think aliasing to old names would be really necessary.

I wonder what you mean by a basic check of the image data? The current status quo attempts to decode the whole image (into a 2d array of pixels) to check for corruption and dimensions. With JuicyPixels gone, those checks are not possible to do (and if it were, adding separate handlers for png/jpeg/webp/gif will be painful). The mime type check was a heuristic that discord.py used for their emoji/sticker/avatar endpoints, which I thought might be good enough.

L0neGamer:
With your answer I'm fine to rubber stamp my approval; I knew nothing of the image checking, so if it's a complex thing that requires juicy pixels then we can forget about it.

aquarial (approved):
good change to reduce deps. Small consistency with other image routes is good

PR #132, opened 2022-06-28
Changed most ToJSON instances to use a new operator

L0neGamer:
This may help with readability.

I made some small other changes as well, mostly just linting hints.

### DIFF
--- a/src/Discord/Internal/Rest/Webhook.hs
+++ b/src/Discord/Internal/Rest/Webhook.hs
@@ -27,14 +27,15 @@ import Discord.Internal.Types
 
 -- aeson introduced type name for json key (text)
 -- https://github.com/haskell/aeson/issues/881
-# if MIN_VERSION_aeson(2, 0, 0)
-import qualified Data.Aeson.Key as Key
-toKey :: T.Text -> Key.Key
-toKey = Key.fromText
-# else
-toKey :: T.Text -> T.Text
-toKey = id
-# endif
+-- this is unneeded due to version bounds.
+-- # if MIN_VERSION_aeson(2, 0, 0)
+-- import qualified Data.Aeson.Key as Key
+-- toKey :: T.Text -> Key.Key
+-- toKey = Key.fromText
+-- # else
+-- toKey :: T.Text -> T.Text
+-- toKey = id
+-- # endif
 
 instance Request (WebhookRequest a) where
   majorRoute = webhookMajorRoute
@@ -84,20 +85,20 @@ data WebhookRequest a where
   --  just use execute webhook.
 
   -- | Returns a previously-sent webhook message from the same token.
-  GetWebhookMessage :: WebhookId 
-                    -> WebhookToken 
-                    -> MessageId 
+  GetWebhookMessage :: WebhookId
+                    -> WebhookToken
+                    -> MessageId
                     -> WebhookRequest Message
   -- | Edits a previously-sent webhook message from the same token.
-  EditWebhookMessage :: WebhookId 
-                     -> WebhookToken 
-                     -> MessageId 
+  EditWebhookMessage :: WebhookId
+                     -> WebhookToken
+                     -> MessageId
                      -> T.Text -- currently we don't support the full range of edits - feel free to PR and fix this
                      -> WebhookRequest Message
   -- | Deletes a previously-sent webhook message from the same token.
-  DeleteWebhookMessage :: WebhookId 
-                       -> WebhookToken 
-                       -> MessageId 
+  DeleteWebhookMessage :: WebhookId
+                       -> WebhookToken
+                       -> MessageId
                        -> WebhookRequest ()
 
 -- | Options for `ModifyWebhook` and `ModifyWebhookWithToken`
@@ -108,10 +109,10 @@ data ModifyWebhookOpts = ModifyWebhookOpts
   } deriving (Show, Read, Eq, Ord)
 
 instance ToJSON ModifyWebhookOpts where
-  toJSON ModifyWebhookOpts{..} = object [(toKey name, val) | (name, Just val) <-
-                         [("channel_id",   toJSON <$> modifyWebhookOptsChannelId),
-                          ("name",   toJSON <$> modifyWebhookOptsName),
-                          ("avatar",  toJSON <$> modifyWebhookOptsAvatar) ] ]
+  toJSON ModifyWebhookOpts{..} = objectFromMaybes
+                         ["channel_id" .=? modifyWebhookOptsChannelId,
+                          "name" .=? modifyWebhookOptsName,
+                          "avatar" .=? modifyWebhookOptsAvatar ]
 
 -- | Options for `CreateWebhook`
 data CreateWebhookOpts = CreateWebhookOpts
@@ -120,9 +121,9 @@ data CreateWebhookOpts = CreateWebhookOpts
   } deriving (Show, Read, Eq, Ord)
 
 instance ToJSON CreateWebhookOpts where
-  toJSON CreateWebhookOpts{..} = object [(name, val) | (name, Just val) <-
-                         [("name",   toJSON <$> Just createWebhookOptsName),
-                          ("avatar",  toJSON <$> createWebhookOptsAvatar) ] ]
+  toJSON CreateWebhookOpts{..} = objectFromMaybes
+                         ["name" .== createWebhookOptsName,
+                          "avatar" .=? createWebhookOptsAvatar ]
 
 -- | Options for `ExecuteWebhookWithToken`
 data ExecuteWebhookWithTokenOpts = ExecuteWebhookWithTokenOpts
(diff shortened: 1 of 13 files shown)


src/Discord/Internal/Rest/Webhook.hs
```diff
@@ -27,14 +27,15 @@ import Discord.Internal.Types
...
-toKey :: T.Text -> Key.Key
-toKey = Key.fromText
-# else
-toKey :: T.Text -> T.Text
-toKey = id
-# endif
+-- this is unneeded due to version bounds.
+-- # if MIN_VERSION_aeson(2, 0, 0)
+-- import qualified Data.Aeson.Key as Key
+-- toKey :: T.Text -> Key.Key
```
aquarial, line 33: Commented out?

src/Discord/Internal/Types/Prelude.hs
```diff
@@ -197,8 +241,17 @@ class Data a => InternalDiscordEnum a where
         | fromIntegral (round i) == i = Just $ round i
         | otherwise = Nothing
 
-toMaybeJSON :: (ToJSON a) => a -> Maybe Value
-toMaybeJSON = return . toJSON
+-- when we upgrade aeson, we need to change Text to Key
+
```
aquarial, line 245: Which version is the problem? Could conditional compilation help>

src/Discord/Internal/Types/Prelude.hs
```diff
@@ -197,8 +241,17 @@ class Data a => InternalDiscordEnum a where
...
+(.==) :: ToJSON a => T.Text -> a -> Maybe Pair
+k .== v = Just (k .= v)
+
+(.=?) :: ToJSON a => T.Text -> Maybe a -> Maybe Pair
+k .=? (Just v) = Just (k .= v)
+_ .=? Nothing = Nothing
+
+objectFromMaybes :: [Maybe Pair] -> Value
+objectFromMaybes = object . catMaybes
```
aquarial, line 254: This is great; `objectFromMaybes` is a well used puzzle piece in this PR

src/Discord/Internal/Types/Prelude.hs
L0neGamer: In my eyes, maintaining this for two separate major versions of something seems like too much of a hassle.

src/Discord/Internal/Rest/Webhook.hs
L0neGamer: Commented out because my formatter kept evaluating it weirdly, and we limit the `aeson` version we're using currently.

src/Discord/Internal/Types/Prelude.hs
aquarial: A fair. We haven't gotten any requests to make the versions looser, so seems like people are happy with it.

It's a static hassle, once the conditional compilation is in place it works.

aquarial: I think the text vs key is the only breaking change Aeson2 makes. We could use the same conditional compilation  `# if MIN_VERSION_aeson(2, 0, 0)` as a template, for the `(.==)` code below. Wouldn't change anything else.

Try implementing the conditional compilation support with aeson 2. If it doesn't work well, I'll merge this PR as is (it's good). If the compilation works then I'd like it in though

aquarial (approved):
I was blocking this because it didn't use conditional compilation for aeson version 2.0, but the cabal version bounds are `< 1.6` right now. Overall better to merge this as is before it gets stale.

Should be no functional changes in this

yutotakano:
I took a look at it, and it might be better to just alias the type instead of introducing `toKey`. Since both `Key.Key` and `T.Text` have IsString instances, no changes to other parts of the code is necessary. Or maybe keep both `toKey` and the alias? 

Here's a patch of what I mean (I didn't know how to submit a PR to a PR? or however it works):

```patch
diff --git discord-haskell.cabal discord-haskell.cabal
index d5fd220..a0e74dd 100644
--- discord-haskell.cabal
+++ discord-haskell.cabal
@@ -89,7 +89,7 @@ library
     , Discord.Internal.Types.ScheduledEvents
   build-depends:
     base ==4.*,
-    aeson >=1.5 && <1.6,
+    aeson >= 1.5 && < 1.6 || >= 2.0 && < 2.2,
     async >=2.2 && <2.3,
     bytestring >=0.10 && <0.11,
     base64-bytestring >=1.1 && <1.2,
diff --git src/Discord/Internal/Rest/Webhook.hs src/Discord/Internal/Rest/Webhook.hs
index ad4b419..7b4a545 100644
--- src/Discord/Internal/Rest/Webhook.hs
+++ src/Discord/Internal/Rest/Webhook.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE CPP #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE RecordWildCards #-}
@@ -25,18 +24,6 @@ import Network.HTTP.Client.MultipartFormData (partBS, partFileRequestBody)
 import Discord.Internal.Rest.Prelude
 import Discord.Internal.Types
 
--- aeson introduced type name for json key (text)
--- https://github.com/haskell/aeson/issues/881
--- this is unneeded due to version bounds.
--- # if MIN_VERSION_aeson(2, 0, 0)
--- import qualified Data.Aeson.Key as Key
--- toKey :: T.Text -> Key.Key
--- toKey = Key.fromText
--- # else
--- toKey :: T.Text -> T.Text
--- toKey = id
--- # endif
-
 instance Request (WebhookRequest a) where
   majorRoute = webhookMajorRoute
   jsonRequest = webhookJsonRequest
@@ -137,7 +124,7 @@ data WebhookContent = WebhookContentText T.Text
                     | WebhookContentEmbeds [CreateEmbed]
   deriving (Show, Read, Eq, Ord)
 
-webhookContentJson :: WebhookContent -> [(T.Text, Value)]
+webhookContentJson :: WebhookContent -> [(AesonKey, Value)]
 webhookContentJson c = case c of
                       WebhookContentText t -> [("content", toJSON t)]
                       WebhookContentFile _ _  -> []
diff --git src/Discord/Internal/Types/Prelude.hs src/Discord/Internal/Types/Prelude.hs
index 4f0b491..20c163e 100644
--- src/Discord/Internal/Types/Prelude.hs
+++ src/Discord/Internal/Types/Prelude.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE MultiParamTypeClasses  #-}
 {-# LANGUAGE FlexibleInstances  #-}
 {-# LANGUAGE RankNTypes  #-}
+{-# LANGUAGE CPP #-}
 
 -- | Provides base types and utility functions needed for modules in Discord.Internal.Types
 module Discord.Internal.Types.Prelude
@@ -45,6 +46,7 @@ module Discord.Internal.Types.Prelude
 
   , (.==)
   , (.=?)
+  , AesonKey
   , objectFromMaybes
   )
 
@@ -65,6 +67,10 @@ import Web.Internal.HttpApiData
 import qualified Data.ByteString as B
 import qualified Data.Text as T
 
+#if MIN_VERSION_aeson(2, 0, 0)
+import qualified Data.Aeson.Key as Key
+#endif
+
 -- | Authorization token for the Discord API
 newtype Auth = Auth T.Text
   deriving (Show, Read, Eq, Ord)
@@ -241,12 +247,21 @@ class Data a => InternalDiscordEnum a where
         | fromIntegral (round i) == i = Just $ round i
         | otherwise = Nothing
 
--- when we upgrade aeson, we need to change Text to Key
-
-(.==) :: ToJSON a => T.Text -> a -> Maybe Pair
+-- Aeson 2.0 uses KeyMaps with a defined Key type for its objects. Aeson up to
+-- 1.5 uses HashMaps with Text for the key. Both types have an IsString instance.
+-- To keep our version bounds as loose as possible while the Haskell ecosystem
+-- (and thus our users) switch over to Aeson 2.0, we use some CPP to define a
+-- AesonKey as an alias.
+#if MIN_VERSION_aeson(2, 0, 0)
+type AesonKey = Key.Key
+# else
+type AesonKey = T.Text
+# endif
+
+(.==) :: ToJSON a => AesonKey -> a -> Maybe Pair
 k .== v = Just (k .= v)
 
-(.=?) :: ToJSON a => T.Text -> Maybe a -> Maybe Pair
+(.=?) :: ToJSON a => AesonKey -> Maybe a -> Maybe Pair
 k .=? (Just v) = Just (k .= v)
 _ .=? Nothing = Nothing

 ```

Annwan:
I think yutotakano's suggestion is the right call, as it allows potential support for both Aeson 1.x and 2.x with next to no change to user code (I think in most cases the only change needed would be to bump the Aeson version in cabal/stack/whatever dependency manager the user uses)

L0neGamer:
You can submit a PR to my repo which can then get merged in here, but that seems messy; I'm gonna apply that patch soon.

aquarial:
Looks great!

Goal is to get this and https://github.com/discord-haskell/discord-haskell/pull/135 into a 1.15 release 

I have a long day ahead, should get around to the release process late tonight

PR #134, opened 2022-07-30
Loosen version bounds for http-client, http-api-data, and req

yutotakano:
This PR loosens the version bounds upwards, to allow a wider range of dependency versions to work with our package.

- `http-client`: Bugfixes and changes in parts we don't use. It's not *really* clear if they follow [PVP](https://pvp.haskell.org/), but going from past changelogs, I think it's fine to expect a major version jump if a future change will break compatibility for the functionality we use.
- `http-api-data`: Drops compatibility for GHC 7, and some changes in parsing. Doesn't affect what we use, and probably won't until at least the next major version.
- `req`: Adds new utility functions and instances, none of which affect our package.

Built and tested with the highest allowed versions for each dependency!

### DIFF
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -96,10 +96,10 @@ library
     containers >=0.6 && <0.7,
     data-default >=0.7 && <0.8,
     emoji ==0.1.*,
-    http-client >=0.6 && <0.7,
+    http-client >=0.6 && <0.8,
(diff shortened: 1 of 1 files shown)


Annwan (approved):
Looks good to me

aquarial:
LGTM

yutotakano:
Thanks both! Merging.

PR #135, opened 2022-08-14
Update the Ready event structure, future-proof Resume Gateway URL

yutotakano:
This PR updates the structure of the `Ready` event dispatched on bot start.

## Changes

- DM channels are no longer sent on the Ready event, so that field has been removed. Instead, for all MESSAGE_CREATE events in DMs, Discord will also send a CHANNEL_CREATE event. discord/discord-api-docs#184
- To future-proof Resume Gateway URLs, we now keep the `resume_gateway_url` received in the Ready event and use that to reconnect if we are performing a Resume. [Discord Changelog for this](https://discord.com/developers/docs/change-log#sessionspecific-gateway-resume-urls)


src/Discord/Internal/Gateway/EventLoop.hs
```diff
@@ -83,12 +90,15 @@ connectionLoop auth intent gatewayHandle log = outerloop LoopStart
     -- | Main connection loop. Catch exceptions and reconnect.
     outerloop :: LoopState -> IO ()
     outerloop state = do
+        -- Read the gateway address, since each new connection to the gateway
+        -- may have a different Host to connect to.
```
aquarial, line 94: I think this comment is unnecesary, because the code is straightforward and you've added good docs on `gatewayHandleHostname` above

src/Discord/Internal/Types/Events.hs
```diff
@@ -213,14 +214,25 @@ reparse val = case parseEither parseJSON $ toJSON val of
                 Left r -> fail r
                 Right b -> pure b
 
+-- | Remove the "wss://" and the trailing slash in a gateway URL, thereby returning
+-- the hostname portion of the URL that we can connect to.
+extractHostname :: String -> HostName
+extractHostname ('w':'s':'s':':':'/':'/':rest) = extractHostname rest
+extractHostname ('/':[]) = []
+extractHostname (a:b) = a:(extractHostname b)
+extractHostname [] = []
```
aquarial, line 223: Cheeky

I can't think of a simpler way without regex, which might not even be simpler.

aquarial (changes requested):
Very quick! 

Thanks for implementing the change @L0neGamer brought attention.

One nitpick, one comment.

src/Discord/Internal/Types/Events.hs
yutotakano: Haha, I considered what the best way to go about this was - especially since Discord doesn't clarify anywhere whether the URL can be assumed to always have the prefix/suffixes either. So I couldn't just do a drop/take too since they might not even be there.

aquarial (approved):

PR #136, opened 2022-08-16
made change to more fully represent the permissions changes

L0neGamer:
the new permissions system needs some sort of representing.

I propose that at some stage, we change from using a `Text` to using a more specified format for perms, but right now this will do so that people can use them at all.


aquarial (approved):
What are the permissions here? There's a few different permissions and overwrites in the repo, it's a bit out of hand. 

I think it's worth adding a markdown file in the /docs folder for permissions.

Overall the changes look good tho

Annwan (approved):
Looks good to me

PR #137, opened 2022-08-16
L10n for application commands

Annwan:
Implements localization for application commands

### DIFF
--- a/src/Discord/Internal/Types/ApplicationCommands.hs
+++ b/src/Discord/Internal/Types/ApplicationCommands.hs
@@ -815,13 +815,10 @@ instance ToJSON LocalizedText where
 instance FromJSON LocalizedText where
   parseJSON = withObject "LocalizedText"
     (\o -> do
-        parsed <- sequenceSecond
-                         . map (second parseJSON) . objectToList $ o
-#if MIN_VERSION_aeson(2, 0, 0)
-        let translations = map (first unKey) parsed
-#else
-        let translations = parsed
-#endif
+        translations <- sequenceSecond
+                        . map (second parseJSON)
+                        . objectToList
+                        $ o
         return $ LocalizedText translations
     ) where
       sequenceSecond :: Monad m => [(a, m b)] -> m [(a, b)]
--- a/src/Discord/Internal/Types/ApplicationCommands.hs
+++ b/src/Discord/Internal/Types/ApplicationCommands.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE CPP #-}
 
 module Discord.Internal.Types.ApplicationCommands
   ( ApplicationCommand (..),
@@ -31,11 +32,15 @@ where
 
 import Data.Aeson (FromJSON (parseJSON), ToJSON (toJSON), Value (Number, Object), object, withArray, withObject, (.!=), (.:), (.:!), (.:?))
 import Data.Aeson.Types (Pair, Parser)
+#if MIN_VERSION_aeson(2, 0, 0)
+  import Data.Aeson.Key(Key(unKey))
+#endif
 import Data.Data (Data)
 import Data.Foldable (Foldable (toList))
 import Data.Scientific (Scientific)
 import qualified Data.Text as T
-import Discord.Internal.Types.Prelude (ApplicationCommandId, ApplicationId, GuildId, InternalDiscordEnum (..), Snowflake, discordTypeParseJSON, objectFromMaybes, (.==), (.=?))
+import Discord.Internal.Types.Prelude (ApplicationCommandId, ApplicationId, GuildId, InternalDiscordEnum (..), Snowflake, discordTypeParseJSON, objectFromMaybes, objectToList, (.==), (.=?))
+import Data.Bifunctor (Bifunctor(second))
 
 type Number = Scientific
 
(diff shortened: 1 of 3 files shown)


src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -696,3 +797,74 @@ instance ToJSON ApplicationCommandPermissions where
         "type" .== applicationCommandPermissionsType,
         "permission" .== applicationCommandPermissionsPermission
       ]
+
+data Locale
+  = Da | De | EnGB | EnUS | EsES | Fr | Hr | It | Lt | Hu | Nl | No | Pl | PtBR
+  | Ro | Fi | SvSE | Vi | Tr | Cs | El | Bg | Ru | Uk | Hi | Th | ZhCN | Ja
+  | ZhTW | Ko
+  deriving (Show, Read, Eq, Ord)
```
L0neGamer, line 805: It might be worth having an "other" option if they add more locales

src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -696,3 +797,74 @@ instance ToJSON ApplicationCommandPermissions where
...
+                               ]
+        let translations = map (\(l, Just t) -> (l, t)) $filter (\x -> (snd x) /= Nothing) maybeTranslations
+        return $ LocalizedText translations
+    ) where
+      sequenceSecond :: Monad m => [(a, m b)] -> m [(a, b)]
+      sequenceSecond l = fmap (zip locales) . sequence $ translations 
+        where
+          unzipped = unzip l
+          locales = fst unzipped
+          translations = snd unzipped
```
L0neGamer, line 870:
```
(locales, translations) = unzip l
```

src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -696,3 +797,74 @@ instance ToJSON ApplicationCommandPermissions where
...
+type TextTranslation = (Locale, T.Text)
+
+newtype LocalizedText = LocalizedText [TextTranslation] deriving (Show, Read, Eq, Ord)
+
+instance ToJSON LocalizedText where
+  toJSON (LocalizedText lt) = object . map (\(l,t) -> (localeString l, toJSON t)) $ lt
+    where
+      localeString l = case l of
+        Da -> "da"
+        De -> "de"
```
L0neGamer, line 816: I really think it would be worth somehow having a bidirectional map for this stuff so that the two locale lists don't get out of sync. alternatively, we just use Texts to represent a locale and don't need us to do any translations

src/Discord/Internal/Types/ApplicationCommands.hs
```diff
@@ -696,3 +797,74 @@ instance ToJSON ApplicationCommandPermissions where
...
+                               , (Hr, "hr"), (It, "it"), (Lt, "lt"), (Hu, "hu")
+                               , (Nl, "nl"), (No, "no"), (Pl, "pl")
+                               , (PtBR, "pt-BR"), (Ro, "ro"), (Fi, "fi")
+                               , (SvSE, "sv-SE"), (Vi, "vi"), (Tr, "tr")
+                               , (Cs, "cs"), (El, "el"), (Bg, "bg"), (Ru, "ru")
+                               , (Uk, "uk"), (Hi, "hi"), (Th, "th")
+                               , (ZhCN, "zh-CN"), (Ja, "ja"), (ZhTW, "zh-TW")
+                               , (Ko, "ko")
+                               ]
+        let translations = map (\(l, Just t) -> (l, t)) $filter (\x -> (snd x) /= Nothing) maybeTranslations
```
L0neGamer, line 862: with `import Data.Maybe` and `{-# LANGUAGE TupleSections #-}, to reduce unsafety and improve readability

```
let translations = mapMaybe (\(l, t) -> fmap (l,) t) $ filter (isNothing . snd) maybeTranslations
```

L0neGamer (commented):
Some recommendations on how to change things

src/Discord/Internal/Types/ApplicationCommands.hs
Annwan: Maybe just going with text might be smarter actually yeah

src/Discord/Internal/Types/ApplicationCommands.hs
L0neGamer: technically what you've written isn't unsafe it's just a bit hard to read

Annwan: I wanted to do it by somehow extracting the Pair list directly from the object, but there is no way to do it while remaining agnostic on the aeson version (spare a bunch of preproc directives

L0neGamer: you can use `AesonKey`, defined in Discord.Types.Prelude?

Annwan: Not really, in Aeson 1.x `Object` is `HashMap AesonKey Value` whereas in Aeson 2.x it is `KeyMap (Map AesonKey Value), and the level of indirection added prevents it from being easily doable

Annwan: Well I could probably throw together an `objectToList :: Object -> [(AesonKey, Value)]` with implementation being dependent on Aeson version

L0neGamer (approved):
Looks overall good to me!

yutotakano (approved):
Looks great to me :)

Annwan:
Well then, all is good, merging

PR #139, opened 2022-08-18
Actually exporting the types for the l10n (and other fixes)

Annwan:
I totally forgot to expose the types at module level (wups)

So uh here are the exports


L0neGamer (approved):
reducing code is great, good job on comments and exports

PR #141, opened 2022-08-28
fix typo I think

Geometer1729:
GetGlobalApplicationCommands was failing with
```
RestCallErrorCode 400 "Library Parse Exception" "restcall - parse exception [Error in $[0]: key \"description_localizations\" not found]
```
This fix worked for me locally. It seems like just a typo / forgetting the field is optional.


Annwan (approved):
Yeah that was a typo on my side, thanks for catching it

PR #143, opened 2022-08-28
Minor Tweaks for backend stuff

L0neGamer:


examples/interaction-commands.hs
```diff
@@ -49,7 +49,7 @@ interactionCommandExample = do
           discordOnEnd = liftIO $ putStrLn "Ended",
           discordOnEvent = eventHandler,
           discordOnLog = \s -> TIO.putStrLn s >> TIO.putStrLn "",
-          discordGatewayIntent = def {gatewayIntentMembers = True, gatewayIntentPrecenses = True}
+          discordGatewayIntent = def {gatewayIntentMembers = True, gatewayIntentPresences = True}
```
yutotakano, line 52: Can you also update line 38 `-- | Replies "pong" to every message that starts with "ping"` with something more accurate? Thanks!

yutotakano (dismissed):
Small comment change requested, but looks good! I'm assuming you've built and ran the new examples.

src/Discord/Internal/Rest/Prelude.hs
```diff
@@ -17,6 +18,9 @@ import Web.Internal.HttpApiData (ToHttpApiData)
 
 import Discord.Internal.Types
 
+import Paths_discord_haskell (version)
```
aquarial, line 21: very neat

discord-haskell.cabal
```diff
@@ -88,7 +151,9 @@ library
     , Discord.Internal.Types.Emoji
     , Discord.Internal.Types.ScheduledEvents
   build-depends:
-    base ==4.*,
+  -- https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history
+  -- below also sets the GHC version effectively. set to == 8.10.* alone for now.
+    base == 4.14.*,
```
yutotakano, line 156: I'm not sure if this is the best idea because our current code builds and runs fine on 9.0.2 as long as base64-bytestring's upper bound is raised.

Maybe that's a better solution? (Not sure about 9.2.4, since there was no Stackage LTS for it, and I didn't want to mess around with my env too much)

examples/ExampleUtils.hs
```diff
@@ -0,0 +1,34 @@
+{-# LANGUAGE LambdaCase #-}
```
yutotakano, line 1: Not all of these look necessary

examples/ExampleUtils.hs
```diff
@@ -0,0 +1,34 @@
...
+import qualified Discord.Requests as R
+import Discord.Types
+import Text.Read (readMaybe)
+
+getToken :: IO T.Text
+getToken = TIO.readFile "./examples/auth-token.secret"
+
+getGuildId :: IO GuildId
+getGuildId = do
+  gids <- readFile "./examples/guildid.secret"
```
yutotakano, line 20: I like the separation of concerns/modules for e.g. how to read the token/gids versus demonstrating how the library works. 👍

yutotakano (commented):
One thing re: version bounds. Looking nice!

I would approve it and just let you change stuff as you want and merge, but I think if you push any new commits it will void my approval anyway. So I'll just leave a plain comment here.

discord-haskell.cabal
L0neGamer: I can relax the bounds by adding `|| == 4.whatever.*` for the 9.0.2 version.

yutotakano: Fair enuff. Ok!

discord-haskell.cabal
```diff
@@ -97,7 +97,7 @@ executable state-counter
                      , discord-haskell
 
 library
-  ghc-options:         -Wall
+  ghc-options:         -Wall -j10
```
Annwan, line 100: `-j10` while speeding up compilation also increases the already significant memory usage of GHC substantially and could lead to issues in memory constrained environments.
So I'm not sure it may be a good idea to use

src/Discord/Internal/Types/Gateway.hs
```diff
@@ -87,23 +88,24 @@ compileGatewayIntent :: GatewayIntent -> Int
...
+                       , (2 ^:  9, gatewayIntentMessageChanges)
+                       , (2 ^: 10, gatewayIntentMessageReactions)
+                       , (2 ^: 11, gatewayIntentMessageTyping)
+                       , (2 ^: 12, gatewayIntentDirectMessageChanges)
+                       , (2 ^: 13, gatewayIntentDirectMessageReactions)
+                       , (2 ^: 14, gatewayIntentDirectMessageTyping)
+                       , (2 ^: 15, gatewayIntentMessageContent)
                        ]
        ]
+  where (^:) :: Int -> Int -> Int = (^)
```
Annwan, line 108: I don't really understand why this was done

stack.yaml
```diff
@@ -2,7 +2,7 @@ extra-package-dbs: []
 packages:
 - '.'
 
-resolver: lts-18.13
+resolver: lts-18.19
```
Annwan, line 5: fyi the latest Stackage lts-18 version is lts-18.27 and discord-haskell works perfectly fine with it

L0neGamer: looks like it's lts-18.28, which I'll try now

Annwan (approved):
Overall looks good to me

PR #148, opened 2022-09-21
Support GHC 9.2.x

matobet:


L0neGamer:
Two things which are minor, one more major: the base bounds can be put back to the simpler less than and greater thans, and the comment should be updated; I'm pretty sure 1.12 time is available; have you tried building this with 9.02,9.2, and 8.10.7?

yutotakano:
8.10.7 is verified by the CI.

matobet:
verified on 9.0.2 and 9.2.4 :+1:

matobet:
Fixed the comment. I also realized that with a few more changes we can support directly up to GHC 9.4.x. I'll leave it up to you how far you want to go at this stage.

L0neGamer (approved):
Assuming you've compiled for 9.4, looks good to me!

matobet:
indeed, verified with 9.4.2 :-)

matobet:
As a side note, maybe it would be worthwile adding a matrix of various GHC versions that we want to support to the CI so it is explicit which GHC versions are supported and we get continuous tests against all of them?

yutotakano:
Sounds like a good idea!

L0neGamer:
Does the stack.yaml resolver not need to be updated for this PR?  
And if you want to add that matrix thing now that sounds good; I don't know what it is.

yutotakano:
> Does the stack.yaml resolver not need to be updated for this PR?  
> And if you want to add that matrix thing now that sounds good; I don't know what it is.

They mean to add more GHC versions to test against in the GitHub actions CI that runs on PRs! It will take a longer time but we're sure we won't be making regressions on versions we can't test locally. :)

yutotakano:
> Does the stack.yaml resolver not need to be updated for this PR?  

I don't think this is necessary -- the stack resolver is there for the default, and for now we can stick to 8.10.7 as the default GHC version that this library uses when you do `stack build`.

I think we're good to merge. I'll update the cabal file on Hackage since updating the metadata doesn't need a version bump.

PR #152, opened 2022-09-30
Permissions

0x3alex:
Bitwise check of Permission for all Roles of an GuildMember or of an specific role.
Bit shifiting of 1 is done like in the discord api. 
Current formatting of the Enum seems to be needed. Check comment.

### DIFF
--- a/src/Discord/Internal/Types/Guild.hs
+++ b/src/Discord/Internal/Types/Guild.hs
@@ -309,13 +309,8 @@ data Permissions =
   deriving (Enum,Show)
 -- | Formatting seems to be like that. fromEnum would otherwise return only 1
 
-
-getHexFromPermission :: Permissions -> Int
-getHexFromPermission p = shift 1 $ fromEnum p
--- | shift index of permission in enum left from 1
-
 hasRolePermission :: Role -> Permissions -> Bool
-hasRolePermission r p = (.&.) (read (T.unpack $ rolePerms r) :: Int) (getHexFromPermission p) > 0
+hasRolePermission r p = (.&.) (read (T.unpack $ rolePerms r) :: Int) (shift 1 $ fromEnum p) > 0
 -- | Check if a given role has the permission
 --   RolePerms need to be an int to be converted into its bits
 --   Bitwise & checks if rolePermission contains the perm
--- a/src/Discord/Internal/Types/Guild.hs
+++ b/src/Discord/Internal/Types/Guild.hs
@@ -366,7 +366,7 @@ hasGuildMemberPermission :: Guild -> GuildMember -> PermissionFlag -> Bool
 hasGuildMemberPermission g gm p = or $ go (memberRoles gm) g
   where
     go [] _ = []
-    go (x:xs) g = case roleIdToRole g x of
+    go (x:xs) _ = case roleIdToRole g x of
                     Nothing ->  [False] <> go xs g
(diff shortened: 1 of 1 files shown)


L0neGamer:
Firstly I think the comments are in the wrong places.

Secondly the data type with each permission should be vertical, although you mentioned having issues with it?

Thirdly, if `roleIdToRole` should either return one value or none, it should probably return Maybe instead of a list.

The "Permissions" data type should probably be called UserPermission or similar; it's a singular permission.

Finally, I dislike that this isn't a generic version, but that can be an extension in future.

I do like the idea of deriving that enum type class and then bit shifting 1 by that value; previous attempts at implementation got stuck in the weeds of how to do this.

src/Discord/Internal/Types/Guild.hs
```diff
@@ -298,6 +301,77 @@ instance FromJSON Role where
          <*> o .: "managed"
          <*> o .: "mentionable"
 
+-- | If there is no such role on the guild return nothing
+--   otherwise return the role. Take the head of the list. List should always be one, because the ID is unique
+roleIdToRole :: Guild -> RoleId -> Maybe Role
+roleIdToRole  g r = bool (Just $ head f) Nothing ( null f)
```
L0neGamer, line 307: Could probably just be `lookup (\x -> roleId x == r) (guildRoles g)`

0x3alex: gonna add it!

0x3alex: lookup expects [(a,b)] gut guildRoles returns [Role]

L0neGamer: `find` is what I was thinking!

0x3alex: implimented it

src/Discord/Internal/Types/Guild.hs
```diff
@@ -298,6 +300,74 @@ instance FromJSON Role where
...
+  | SEND_MESSAGES_IN_THREADS
+  | USE_EMBEDDED_ACTIVITIES
+  deriving (Enum,Show)
+
+
+-- | Check if a given role has the permission
+--   RolePerms need to be an int to be converted into its bits
+--   Bitwise & checks if rolePermission contains the perm
+hasRolePermission :: Role -> PermissionFlag -> Bool
+hasRolePermission r p = (.&.) (read (T.unpack $ rolePerms r) :: Int) (shift 1 $ fromEnum p) > 0
```
L0neGamer, line 356: Use of `read` is not recommended because it'll just error if it fails. Maybe we can use `readMaybe`?

src/Discord/Internal/Types/Guild.hs
```diff
@@ -298,6 +302,75 @@ instance FromJSON Role where
...
+
+
+
+-- | Check if any Role of an GuildMember has the needed permission
+--   If the result of roleIdToRole is Nothing, it appends an "False"
+--   Otherwise it checks for the needed permission
+hasGuildMemberPermission :: Guild -> GuildMember -> PermissionFlag -> Bool
+hasGuildMemberPermission g gm p = or $ go (memberRoles gm) g
+  where
+    go [] _ = []
```
L0neGamer, line 368: Another minor thing: this function doesn't need the second parameter. As a minor thing you could encode the or'ing yourself in this function but that's a style thing honestly

L0neGamer (approved):
I'm happy for this to become part of discord-haskell, pending ci

PR #153, opened 2022-10-02
Add CI build checks for GHC 8.10.7, 9.2, and 9.4

yutotakano:
While we currently have a build check for the main supported version (Stackage lts-18.28 = GHC 8.10.7), our library should work with GHC 9.2 and 9.4 as well. In fact, recent pull requests have been explicitly designed to support the three versions. This PR introduces continuous integration for those.

It is more complex than it needs to be, since I have made it so the GHC versions to build against are dynamically parsed from the cabal file's `tested-with` field. Future changes to supported versions will likely require minimal changes to the CI files.

All comment-worthy things have been commented!

~~**One thing I want opinions for:** I have added build checks for both cabal and stack for each version, resulting in 6 builds. Is this unnecessary? If so, should we keep the stack builds or the cabal builds? For reference, the one we currently have is a stack build.~~

EDIT: as discussed on Discord, our development cycle isn't very fast that this is really problematic from a time point of view. Further, concerns about cache sizes are dismissable too since GitHub has a limit of 10GB and our cache is about 2-3 GB total. This is ready for merge as-is pending review.

The time it takes for the six builds in total (parallel) is roughly 4-5 minutes for a dependency-cached rebuild, and 25-30 minutes for a full recompilation of all dependencies.


L0neGamer (approved):
Well worth having and well made!

PR #154, opened 2022-10-08
Expose Discord.Internal.Rest.ScheduledEvents

1Computer1:
Looks like this was implemented a while ago and the corresponding Types module was exposed, but not the Rest one.


L0neGamer (approved):
Exporting this stuff is definitely useful

yutotakano (approved):

PR #155, opened 2022-10-16
Select options

L0neGamer:


Annwan (approved):
LGTM

src/Discord/Internal/Types/Interactions.hs
```diff
@@ -255,10 +255,34 @@ instance FromJSON ComponentData where
             2 -> return $ ButtonData cid
             _ | t `elem` [3, 5, 6, 7, 8] ->
               SelectMenuData cid
-                <$> v .: "values"
-            _ -> fail "unknown interaction data component type"
+                <$> parseJSON (toJSON v)
```
yutotakano, line 258: This seems cursed, is there no cleaner way to go about this?

yutotakano:
LGTM except one comment which I'd like an explanation of, although in my humble opinion having not done a lot around this, the whole interaction-related stuff is getting unwieldy and some nicer types (ala #156) or some shorter smart-constructors/DSL could be due in the future.

src/Discord/Internal/Types/Interactions.hs
L0neGamer: Are you talking about the icky "type in list" or the ToJSON parsejson?

yutotakano: The latter! I could just be misunderstanding the types here but we're essentially making a pure value into an Aeson Value, then parsing that back to a pure value no?

L0neGamer: Ah, so the `v` at this point is an `Object`, which is an aesom construct which we can query for fields. However, we want to parse the object knowing it's a select menu, so we have to turn it back into a `Value` before we parse it.

L0neGamer: It's the same thing that we initially do to get the internal events, just wrapped up in the function `reparse`.

yutotakano: Ahhhhh yes, I see what it does now. It's an artifact of using sub-ADTs for the various different shapes the JSON can be. Makes sense, all good to go!

yutotakano (approved):

Issue #156, opened 2022-10-19
Move to unique data types within sum types; remove partial functions in records


L0neGamer:
Currently the code base is rife with records that are partial functions. Additionally, it's hard for users of the library to extract and separate processing of different elements of sum types, meaning they either have to do all the processing of each sum type in the case/function they're writing, or they need to have many "dead" cases where, say, an Event is passed in but they only want to process Ready events, so a maybe value is returned.

This would be a large refactor, and should likely be done in parts.

L0neGamer:
Another method may be to create typeclasses for things like interactions, but this may not work for things like recieving an arbitrary interaction. It would work well with sending interactions, I think.

tam-carre:
Could the `DataKinds` extension be useful here to automatically promote data constructors to type constructors?

L0neGamer:
I'm not sure how datakinds would apply here. Do you have a minimal example in mind?

L0neGamer:
This might be similar to what I'm envisaging: https://hackage.haskell.org/package/open-union. Although I guess it doesn't allow for direct pattern matching, it might allow for that function style one.

L0neGamer:
I think I'm misunderstanding myself here. We can have different solutions to sending and recieving data due to the fact that there's never a case (that I recall) that we send full application command, interaction, channel, etc objects; we usually send objects that are similar, yes, but constructed separately.

tam-carre:
Sorry for the noise, I think I have completely misunderstood what DataKinds does. I had read that it "promotes data constructors to type constructors" but what I didn't understand is that those type constructors don't return a type of kind `*` which can have values.

L0neGamer:
I think one thing that might be useful is using JSON's Alternative instance. This could help with the weirdness seen in 155 with the type integers.

PR #158, opened 2022-10-29
Cache Changes

L0neGamer:
making the cache a bit more reliable and useful, while also letting people not use it if they don't wish.

There are still many more things that can and probably should be cached, like threads, but I want to wait for the better types system I mention in #156.


Annwan (approved):
Aside from a tiny formatting nitpick that you are absolutely free to ignore, looks good to me

PR #159, opened 2022-10-29
Add simple interaction commands example

tam-carre:
Add a simpler example of Slash Command management.

In contrast with the old example, this also unregisters outdated slash commands.

Currently, this example does not create any option to keep it simple, but maybe this would be desirable?


L0neGamer:
Would it be possible to use the "default test server id" thing used in some of the other files? Means we can avoid registering global chat inputs, which take a while to register for test inputs.

tam-carre:
@L0neGamer No problem: 3f45fae

As an aside, I'm not sure what determines how much time global commands take to register, but it's been pretty much instant on my bot that's on only two guilds (if that's even a factor).

L0neGamer:
Global commands are meant to take an hour to register or so.

Could you also add this example to the .cabal file?

tam-carre:
@L0neGamer cec53cb

There also were some type errors that I fixed

L0neGamer (approved):
Looks good to me!

Issue #163, opened 2023-01-29
permissions in Role should probably be represented with an Integral type


Annwan:
Currently `rolePerms` in `Role` are of type `Data.Text.Text` but it actually is an integer representing a bitset, it would therefore make more sense to have it be an Integral type, it would also simplify permission calculations.

L0neGamer:
Ideally we would have a type that it's easy to request permissions from without the library user having to have the bits required to hand

XanderDJ:
I don't mind doing this

Annwan:
uh something like
```hs
newtype Permission = Permission
    { permissionCreateInstantInvite :: Bool
    , permissionKickMembers :: Bool
    , permissionBanMembers :: Bool
    , permissionAdministrator :: Bool
    , -- ...
    }
```?

**Comment by XanderDJ** [comment]
I was more so thinking like:
`
data Permissions = CreateInstantInvite | ...

hasPermissions :: [Permission] -> (type RolePerms = Int) -> Bool
`
You could also check every permission to see if it is set and then make that type.

**Comment by Annwan** [comment]
That could work too, the only problem I might see with that approach is when setting permissions
If you want to update a permission with the record approach you just have to do something like that `newPerms = oldPerms {permissionSendMessage = False}` instead of having to filter the list (for deletion) or check against duplicates (when authorizing)

**Comment by XanderDJ** [comment]
```haskell
addPermissions, removePermissions :: [Permission] -> RolePerms -> RolePerms
newPermissions :: [Permission] -> RolePerms
```

I don't think that's as big of a problem if you just make sure the functions are there.

L0neGamer:
Alternatively, we store a fixed size int for the permissions, then give functions which get the permissions from it and to it. This means that if they add permissions and we don't update, people can still access those permissions if they need.

XanderDJ:
Well in my example RolePerms would just be a type alias for an integer so that would already be a thing.

L0neGamer:
I think newtype would be preferred but yeah that sounds decent.

Annwan:
yeah that'd work then
we'd just need to add
```hs
setPermission :: Permission -> RolePerms -> RolePerms
clearPermission :: Permission -> RolePerms -> RolePerms
```
also to be consitent with the way wrapper types are done in the lib it would be something more like
```hs
newtype RolePerms = RolePerms Integer
```

XanderDJ:
Yea I get the newtype. I will get this done monday if that is fine.

[...]

PR #164, opened 2023-01-30
Finished implementation of new Role Permissions

XanderDJ:
As described in Issue #163 I've implemented a new RolePermissions system that expanded the previous one present in Guild.hs. Added as an internal type to Types and then exported forwards so taht it's available when importing Discord.Types. New representation of RolePermissions through a newtype that wraps around an Integer. The other functions are all bit operations mostly.

### DIFF
--- a/src/Discord/Internal/Rest/Guild.hs
+++ b/src/Discord/Internal/Rest/Guild.hs
@@ -177,7 +177,7 @@ instance ToJSON CreateGuildBanOpts where
 -- | Options for `ModifyGuildRole`
 data ModifyGuildRoleOpts = ModifyGuildRoleOpts
   { modifyGuildRoleOptsName            :: Maybe T.Text
-  , modifyGuildRoleOptsPermissions     :: Maybe T.Text
+  , modifyGuildRoleOptsPermissions     :: Maybe RolePermissions
   , modifyGuildRoleOptsColor           :: Maybe DiscordColor
   , modifyGuildRoleOptsSeparateSidebar :: Maybe Bool
   , modifyGuildRoleOptsMentionable     :: Maybe Bool
@@ -186,7 +186,7 @@ data ModifyGuildRoleOpts = ModifyGuildRoleOpts
 instance ToJSON ModifyGuildRoleOpts where
   toJSON ModifyGuildRoleOpts{..} = objectFromMaybes
                        ["name" .=? modifyGuildRoleOptsName,
-                        "permissions" .=? modifyGuildRoleOptsPermissions,
+                        "permissions" .=? fmap (T.pack . show) modifyGuildRoleOptsPermissions,
                         "color" .=? modifyGuildRoleOptsColor,
                         "hoist" .=? modifyGuildRoleOptsSeparateSidebar,
                         "mentionable" .=? modifyGuildRoleOptsMentionable]
--- a/src/Discord/Internal/Rest/Guild.hs
+++ b/src/Discord/Internal/Rest/Guild.hs
@@ -186,7 +186,7 @@ data ModifyGuildRoleOpts = ModifyGuildRoleOpts
 instance ToJSON ModifyGuildRoleOpts where
   toJSON ModifyGuildRoleOpts{..} = objectFromMaybes
                        ["name" .=? modifyGuildRoleOptsName,
-                        "permissions" .=? fmap (T.pack . show) modifyGuildRoleOptsPermissions,
+                        "permissions" .=? modifyGuildRoleOptsPermissions,
                         "color" .=? modifyGuildRoleOptsColor,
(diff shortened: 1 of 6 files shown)


XanderDJ:
Also fixed random stuff I've encountered here and there like typos

L0neGamer:
I said I would review in the morning but I have some comments now.
You create a map from an enum to that enum and its bit representation, and then use the theoretically unsafe M.! to access it. Why not just create a function that does the same thing? This would remove the need for a map constant in code, as well as reduce the requirement of M.!.

I have more comments but they can wait for tomorrow.

XanderDJ:
> I said I would review in the morning but I have some comments now. You create a map from an enum to that enum and its bit representation, and then use the theoretically unsafe M.! to access it. Why not just create a function that does the same thing? This would remove the need for a map constant in code, as well as reduce the requirement of M.!.
> 
I had already changed this since I've refactored it a bit. I squashed the commits hence why it's only 1 commit

src/Discord/Internal/Rest/Guild.hs
```diff
@@ -186,7 +186,7 @@ data ModifyGuildRoleOpts = ModifyGuildRoleOpts
 instance ToJSON ModifyGuildRoleOpts where
   toJSON ModifyGuildRoleOpts{..} = objectFromMaybes
                        ["name" .=? modifyGuildRoleOptsName,
-                        "permissions" .=? modifyGuildRoleOptsPermissions,
+                        "permissions" .=? fmap (T.pack . show) modifyGuildRoleOptsPermissions,
```
L0neGamer, line 189: This indicates that we may want to create a ToJSON instance for RolePermissions

src/Discord/Internal/Types/Guild.hs
```diff
@@ -276,78 +274,16 @@ instance FromJSON Role where
          <*> o .: "color"
          <*> o .: "hoist"
          <*> o .: "position"
-         <*> o .: "permissions"
+         <*> fmap (readMaybe . T.unpack) (o .: "permissions") .!= RolePermissions 0
```
L0neGamer, line 277: the use of readmaybe here indicates that we may want a FromJSON instance for RolePermissions. Further, if we can't read the string they throw at us, we should fail, so the use of readMaybe here isn't advised.

src/Discord/Internal/Types/RolePermissions.hs
```diff
@@ -0,0 +1,119 @@
...
+-- | Check if any Role of an GuildMember has the needed permission
+--   If the result of roleIdToRole is Nothing, it appends an "False"
+--   Otherwise it checks for the needed permission
+hasGuildMemberPermission :: Guild -> GuildMember -> PermissionFlag -> Bool
+hasGuildMemberPermission g gm p = or $ go (memberRoles gm)
+  where
+    go [] = []
+    go (x : xs) = case roleIdToRole g x of
+      Nothing -> [False] <> go xs
+      Just a -> [p `hasRolePermission` rolePerms a] <> go xs
```
L0neGamer, line 119: minor: newline at end of file please!

L0neGamer, line 119: Why not something like
```haskell
go [] = False
go (x:xs) = case roleIdToRole g of
    Nothing -> go xs
    Just a -> (p `hasRolePermission` rolePerms a) || go xs
```

Then we don't need to try and construct a list and then or through it.

Alternatively, the lists could be constructed by prepending the elements you're suggesting to the front.

src/Discord/Internal/Types/RolePermissions.hs
```diff
@@ -0,0 +1,119 @@
...
+
+-- | Remove a permission from a user by clearing the bit
+clearRolePermission :: PermissionFlag -> RolePermissions -> RolePermissions
+clearRolePermission p = (.&.) (complement . permissionBits $ p)
+
+combinePermissions :: [PermissionFlag] -> RolePermissions
+combinePermissions = foldr ((.|.) . permissionBits) 0
+
+-- | Check if any Role of an GuildMember has the needed permission
+--   If the result of roleIdToRole is Nothing, it appends an "False"
```
L0neGamer, line 111: > it appends an "False"

to

> it prepends a `False`

L0neGamer (changes requested):
I'm generally happy that this will work as intended, but I have a couple of comments I'd like addressed first.

src/Discord/Internal/Types/RolePermissions.hs
XanderDJ: This function was one of aquarials one, I didn't want to touch it tbh I just changed it so that it worksw with the new types and that's basically it. I'll implement the feedback later

XanderDJ:
Implemented all feedback

L0neGamer (approved):

PR #170, opened 2023-04-30
Latency measure

L0neGamer:
Add some utilities to measure latency

### DIFF
--- a/src/Discord/Internal/Gateway/EventLoop.hs
+++ b/src/Discord/Internal/Gateway/EventLoop.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TupleSections #-}
 
 -- | Provides logic code for interacting with the Discord websocket
 --   gateway. Realistically, this is probably lower level than most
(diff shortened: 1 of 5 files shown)


src/Discord/Internal/Gateway/EventLoop.hs
```diff
@@ -182,11 +188,14 @@ runEventLoop thehandle sendablesData log = do loop
...
-                                        writeChan (librarySendables sendablesData) (Heartbeat sq)
+                                        sendHeartbeat sendablesData (gatewayHandleHeartbeatTimes thehandle) sq
                                         loop
       Right (InvalidSession retry) -> pure $ if retry then LoopReconnect else LoopStart
       Right Reconnect        -> pure LoopReconnect
-      Right HeartbeatAck     -> loop
+      Right HeartbeatAck     -> do
+        currTime <- getCurrentTime
+        result <- atomicModifyIORef' (gatewayHandleHeartbeatAckTimes thehandle) (dupe . const currTime)
+        result `seq` loop
```
yutotakano, line 198: Doesn't `atomicModifyIORef'` already force the result, so we don't need the `seq`? I'm going off the docs for `atomicallyModifyIORef'` here, please correct me if wrong.

L0neGamer: it seq's whatever is put into the ioref with what it outputs, so it needs this output to be evaluated as well

L0neGamer: This is an incorrect understanding, and I can adjust the functions as a result; let me know if there's anything else to change!

yutotakano: LGTM, but I can't do a local test for reasons, I trust you fully that it's fine though.

yutotakano (approved):

PR #173, opened 2023-06-14
Global names

Gregory1234:
Discord has recently made a change to the username system, which added global nicknames.
This change exposes the new "global_name" field from the API.
https://discord.com/developers/docs/resources/user#user-object-user-structure


Gregory1234:
Can it be merged then?

L0neGamer (approved):
Lgtm, passes compilation. I'll ask Karl how his sharding is going and if it's close this will go into a new major with that, else I'll look into a new minor with just this.

Issue #175, opened 2023-07-19
GHC 9.6 support


penelopeysm:
Hello!

I'm wondering what it would take for GHC 9.6 support (`base == 4.18.0.0`) to be added. The following are the minimal changes needed to get the `discord-haskell` package to build successfully; however, I haven't tested it in any way (e.g. by running the executables).

Do you have a standard procedure for bumping dependencies and testing? Or is this enough? I'm happy to look into this further and make a PR (unless it gets too involved).

```diff
diff --git a/discord-haskell.cabal b/discord-haskell.cabal
index 2e7dcec..71ebe48 100644
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -1,6 +1,6 @@
 cabal-version:       2.0
 name:                discord-haskell
-version:             1.15.5
+version:             1.16.0
 description:         Functions and data types to write discord bots.
                      Official discord docs <https://discord.com/developers/docs/reference>.
                      .
@@ -159,8 +159,8 @@ library
   build-depends:
   -- https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history
   -- below also sets the GHC version effectively. set to == 8.10.*, == 9.0.*., == 9.2.*, == 9.4.*
-    base == 4.14.* || == 4.15.* || == 4.16.* || == 4.17.*,
-    aeson >= 1.5 && < 1.6 || >= 2.0 && < 2.2,
+    base == 4.14.* || == 4.15.* || == 4.16.* || == 4.17.* || == 4.18.*,
+    aeson >= 1.5 && < 1.6 || >= 2.0 && < 2.3,
     async >=2.2 && <2.3,
     bytestring >=0.10 && <0.12,
     base64-bytestring >=1.1 && <1.3,
@@ -169,7 +169,7 @@ library
     emoji ==0.1.*,
     http-client >=0.6 && <0.8,
     iso8601-time >=0.1 && <0.2,
-    MonadRandom >=0.5 && <0.6,
+    MonadRandom >=0.5 && <0.7,
     req >=3.9 && <3.14,
     safe-exceptions >=0.1 && <0.2,
     text >=1.2 && <3,
@@ -177,7 +177,7 @@ library
     websockets >=0.12 && <0.13,
     network >=3.0.0.0 && <3.2.0.0,
     wuss >=1.1 && <3,
-    mtl >=2.2 && <2.3,
+    mtl >=2.2 && <2.4,
[...]

PR #176, opened 2023-07-19
Update dependencies for GHC 9.6

penelopeysm:
cf. #175

### DIFF
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -19,6 +19,7 @@ build-type:          Simple
 tested-with:         GHC == 8.10.7
                    , GHC == 9.2
                    , GHC == 9.4
+                   , GHC == 9.6
 extra-doc-files:     README.md
                    , changelog.md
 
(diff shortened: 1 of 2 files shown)


yutotakano (approved):

yutotakano:
Unrelated but a future issue: I noticed that while Cabal CI works fine across versions, the Stack CI build check always uses 8.10.7 across all of the build matrix - so it's kind of useless. This is because it uses the stack.yaml file which specifies the stackage snapshot which ties the GHC version.

To solve this we need to create different stack.yaml files for each GHC version we support, using a different snapshot for each. Then pass that to CI with a --stack-yaml flag.

Alternatively, my subjective opinion but Stack seems to be falling out of use in favour of directly using cabal, so perhaps we can cut that out of CI entirely and focus on cabal.

penelopeysm:
(I'm not sure what my opinion counts for here, as I mostly lurk around the Haskell community!) Stack usage has been consistently dropping over time, this is reflected in the annual Haskell survey results — compare, say, [2022](https://taylor.fausak.me/2022/11/18/haskell-survey-results/#s3q0) with [2019](https://taylor.fausak.me/2019/11/16/haskell-survey-results/#s3). But at the same time, it seems that there's still a significant proportion of people using Stack (49% in 2022). Given this, and since Stack support doesn't seem too difficult to do correctly, I'd be open to picking it up in a separate PR 🙂

L0neGamer:
This PR looks good, I'll merge it in the morning.

Re: stack automation, I'm happy for us to move in either direction, but the current position isn't good. I think reducing the stack stuff we have to support would be nice.

L0neGamer (approved):

PR #181, opened 2023-07-25
Bump emojis dependency to 0.1.3

penelopeysm:
Hi, just a quick follow-up on #179 — my upstream PR to `emojis` got merged (https://github.com/jgm/emojis/pull/3), so bumping the dependency version here will give us access to the full set of emoji that Discord supports :)

### DIFF
--- a/discord-haskell.cabal
+++ b/discord-haskell.cabal
@@ -167,7 +167,7 @@ library
     base64-bytestring >=1.1 && <1.3,
     containers >=0.6 && <0.7,
     data-default >=0.7 && <0.8,
-    emojis >=0.1.2 && <0.2,
+    emojis >=0.1.3 && <0.2,
     http-client >=0.6 && <0.8,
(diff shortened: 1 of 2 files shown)


L0neGamer:
Looks like the stack ci build is unhappy. I'll ask another maintainer to rubber stamp my pr that removes it and then try rerun the pipelines

penelopeysm:
Oh, of course, it wouldn't be on Stackage. Forgot about that sorry! but if you all are going down the remove-Stack path I guess it won't be an issue.

L0neGamer:
I think we're maintaining having a stack.yaml so that contributors can build with stack if they so wish. Would you be able to add emojis-0.1.3 to the stack.yaml?

penelopeysm:
I should have poked one of you about this, I must've forgot, sorry. Anyway, it should be good to merge now; I've tested with `stack build` on Stack 2.9.3.

L0neGamer (approved):

PR #183, opened 2023-08-02
Only perform one Stack build since build matrix is useless for it

### LINKED ISSUE #177: [Discussion] Stack support
Creating a separate thread to discuss the questions about stack support raised in #176 

## Previous comments

> Unrelated but a future issue: I noticed that while Cabal CI works fine across versions, the Stack CI build check always uses 8.10.7 across all of the build matrix - so it's kind of useless. This is because it uses the stack.yaml file which specifies the stackage snapshot which ties the GHC version.
>
> To solve this we need to create different stack.yaml files for each GHC version we support, using a different snapshot for each. Then pass that to CI with a --stack-yaml flag.
>
> Alternatively, my subjective opinion but Stack seems to be falling out of use in favour of directly using cabal, so perhaps we can cut that out of CI entirely and focus on cabal.

_Originally posted by @yutotakano in https://github.com/discord-haskell/discord-haskell/issues/176#issuecomment-1642580138_

> (I'm not sure what my opinion counts for here, as I mostly just lurk in the Haskell community!) Stack usage has been consistently dropping over time, this is reflected in the annual Haskell survey results — compare, say, [2022](https://taylor.fausak.me/2022/11/18/haskell-survey-results/#s3q0) with [2019](https://taylor.fausak.me/2019/11/16/haskell-survey-results/#s3). But at the same time, it seems that there's still a significant proportion of people using Stack (49% in 2022). Given this, and since Stack support doesn't seem too difficult to do correctly, I'd be open to picking it up in a separate PR 🙂

_Originally posted by @penelopeysm in https://github.com/discord-haskell/discord-haskell/issues/176#issuecomment-1642815557_

>Re: stack automation, I'm happy for us to move in either direction, but the current position isn't good. I think reducing the stack stuff we have to support would be nice.

_Originally posted by @L0neGamer in https://github.com/discord-haskell/discord-haskell/issues/176#issuecomment-1642881615_

yutotakano:
Fixes #177 , by moving the stack build outside of the GHC version matrix and only building it once using the pinned GHC version specified in stack.yaml (indirectly through the snapshot version).

This doesn't reduce any CI test coverage compared to what we have currently, but save on runner resources since we don't execute the same Stack build multiple times.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,7 +37,7 @@ jobs:
         # quotes on the outside than try to escape the inner double quotes.
         run: echo 'versions={"ghc-version":${{ steps.parse.outputs.tested-with-versions }}}' >> $GITHUB_OUTPUT
 
-  build:
+  build-cabal:
     name: Cabal Build Check
     needs: generate-matrix
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,14 +28,14 @@ jobs:
       - name: Parse the cabal tested-with stanza
         id: parse
         run: |
-          echo "::set-output name=tested-with-versions::$(runghc .github/workflows/parseVersions.hs)"
+          echo "tested-with-versions=$(runghc .github/workflows/parseVersions.hs)" >> $GITHUB_OUTPUT
 
       - name: Set the GHC matrix for the next job
         id: set-ghc-matrix
         # We use single quotes here, since the output from the previous step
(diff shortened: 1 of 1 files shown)


yutotakano:
The behaviour of the new CI can be verified by checking this PR's checks (note in particular, the Stack Build Check job, which is separate from all the Cabal ones).

L0neGamer (approved):
Looks good to me!

yutotakano:
Thanks!

PR #187, opened 2023-09-06
Fix Activity Buttons

L0neGamer:


src/Discord/Internal/Types/Guild.hs
```diff
@@ -1,21 +1,32 @@
...
 import Discord.Internal.Types.Emoji (Emoji, StickerItem)
-import Data.List
+
+-- we support aesons before and after 2, so we have to play around with this
+-- to make sure we import the right module
+#if MIN_VERSION_aeson(2, 0, 0)
+import qualified Data.Aeson.KeyMap as KM
+#else
+import qualified Data.HashMap.Strict as KM
+#endif
```
aquarial, line 30: Reminder for discussions about aeson < 2.0. Might not need this

src/Discord/Internal/Types/Guild.hs
```diff
@@ -119,26 +130,42 @@ instance FromJSON GuildUnavailable where
...
+    parseIncomingActivity :: Value -> Parser (Activity, Maybe [T.Text])
+    parseIncomingActivity = withObject "PI Activity w/ BtnNames" $ \o -> do
+      let o' = KM.delete "buttons" o
+      act <- parseJSON (Object o')
+      buttonNames <- o .:? "buttons"
+      pure (act, buttonNames)
 
 -- | Object for a single activity
 --
 -- https://discord.com/developers/docs/topics/gateway#activity-object
```
aquarial, line 163: Minor note: the docs changed and this url should be changed to `https://discord.com/developers/docs/topics/gateway-events#activity-object`

src/Discord/Internal/Types/Guild.hs
```diff
@@ -119,26 +130,42 @@ instance FromJSON GuildUnavailable where
        GuildUnavailable <$> o .: "id"
 
 data PresenceInfo = PresenceInfo
-  { presenceUserId     :: UserId
+  { presenceUserId     :: Maybe UserId
```
aquarial, line 133: Is the UserID optional now? As in, are we running into crashes because the current code assumes the id is always present? if so merging sooner is better than any nitpicks in this review

aquarial (commented):
Good changes here!

I'm grateful you're keeping up with the activities/buttons changes here

aquarial (approved):
LGTM

PR #188, opened 2023-09-08
Upgrade resolver and some other clean ups

L0neGamer:
Happy to drop commits from this, but I wanted to take the opportunity while I was upping the resolver and moving on from pre aeson-2 to do a little bit of clean up for some things I noticed.

### DIFF
--- a/src/Discord/Internal/Types/Channel.hs
+++ b/src/Discord/Internal/Types/Channel.hs
@@ -170,7 +170,7 @@ instance FromJSON Channel where
     case type' of
       0 ->
         ChannelText  <$> o .:  "id"
-                     <*> o .:? "guild_id" .!= 0
+                     <*> o .:  "guild_id"
                      <*> o .:  "name"
                      <*> o .:  "position"
                      <*> o .:  "permission_overwrites"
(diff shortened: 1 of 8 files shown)


src/Discord/Internal/Types/Channel.hs
```diff
@@ -170,7 +170,7 @@ instance FromJSON Channel where
     case type' of
       0 ->
         ChannelText  <$> o .:  "id"
-                     <*> o .:? "guild_id" .!= 0
```
aquarial, line 173: Good idea to remove the default handling. A missing guild_id should be a crash

src/Discord/Internal/Types/Prelude.hs
```diff
@@ -301,15 +285,15 @@ objectFromMaybes = object . catMaybes
 --
 -- Public creation of this datatype should be done using the relevant smart
 -- constructors for Emoji, Sticker, or Avatar.
-data Base64Image a = Base64Image T.Text T.Text
+data Base64Image a = Base64Image { mimeType :: T.Text, base64Data :: B.StrictByteString }
```
aquarial, line 288: B.StrictByteString doesn't exist in cabal 8.10 (generating a github actions build error). It's just a type alias for B.ByteString (TIL). 

Should probably use B.ByteString for compatibility?

L0neGamer: Made this adjustment. Odd that the names are like that, yeah. Personally I'd make StrictByteString the actual name so it's easier to tell it and LazyByteString apart, but I'm not a maintainer there...

aquarial (approved):
Looks good! 



Let's drop Aeson <2

PR #189, opened 2023-09-09
Cache erroring

L0neGamer:
Make the cache slightly more explicit in its two failure states so that users are less likely to get confused why they can't access it.

### DIFF
--- a/docs/cache.md
+++ b/docs/cache.md
@@ -1,7 +1,5 @@
 ### Cache
 
-The cache (`readCache`) is currently deprecated.
-
-It's capable of working, but the code to update is not written.
+The cache (`readCache`) is currently a bit odd at times.
 
 Current source code is at [Discord.Internal.Gateway.Cache](../src/Discord/Internal/Gateway/Cache.hs)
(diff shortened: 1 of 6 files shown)


aquarial:
Thinking about another way to approach this:

The cache uses [two parts from the `InternalReady` event right now](https://github.com/discord-haskell/discord-haskell/blob/617fa38cdba981ff3c7ffb8df429da429045eda3/src/Discord/Internal/Gateway/Cache.hs#L35-L36): the User object, and a PartialApplication.

Both of those have a REST endpoint (https://discord.com/developers/docs/resources/user#get-user https://discord.com/developers/docs/resources/application#get-current-application)

The code already [runs a REST request for GetCurrentUser](https://github.com/discord-haskell/discord-haskell/blob/617fa38cdba981ff3c7ffb8df429da429045eda3/src/Discord.hs#L119) before running discordOnStart.

We can run both these rest requests and populate the cache at the beginning! The external interface would remain the same because the cache would always be populated and now it's safe to lookup during the onStart handler

aquarial:
Here''s an implementation: https://github.com/discord-haskell/discord-haskell/pull/190

Issue #186, opened 2023-09-01
Improve error messages of unfilled cache


L0neGamer:
In #185 it was unclear what exactly was wrong with using the cache without knowledge of library internals. 

We should improve the cache by using a tvar to ensure it's always filled, and initially filling it with an error that says "cache is either disabled, or is accessed before the Ready event has been received".

Alternative methods that don't use error would be appreciated, but as this is a value that once set should stay set I think I prefer this implementation.

An alternative could be a custom type and cache accessor that throws an exception in the usual way when the cache is not available.

aquarial:
Can this issue be closed? 

PR #190 means the cache won't be empty outside of crashes.]

TVar vs MVar is an interesting question. I like the MVar situation because the invariants are simple for internal code and a user has to go out of their way to take the cache out of the MVar.

PR #190, opened 2023-09-10
Populate cache from rest

aquarial:
Populates the cache using the rest requests before the onStart handler

### DIFF
--- a/src/Discord.hs
+++ b/src/Discord.hs
@@ -42,6 +42,7 @@ import Discord.Handle
 import Discord.Internal.Rest
 import Discord.Internal.Rest.User (UserRequest(GetCurrentUser))
 import Discord.Internal.Gateway
+import qualified Discord.Requests as R
 
 -- | A `ReaderT` wrapper around `DiscordHandle` and `IO`. Most functions act in
 -- this monad
@@ -116,18 +117,29 @@ runDiscord opts = do
 -- | Runs the main loop 
 runDiscordLoop :: DiscordHandle -> RunDiscordOpts -> IO T.Text
 runDiscordLoop handle opts = do
-  resp <- liftIO $ writeRestCall (discordHandleRestChan handle) GetCurrentUser
+  resp <- startupRestCalls
   case resp of
     Left (RestCallInternalErrorCode c e1 e2) -> libError $
              "HTTP Error Code " <> T.pack (show c) <> " " <> TE.decodeUtf8 e1
                                                    <> " " <> TE.decodeUtf8 e2
     Left (RestCallInternalHttpException e) -> libError ("HTTP Exception -  " <> T.pack (show e))
     Left (RestCallInternalNoParse _ _) -> libError "Couldn't parse GetCurrentUser"
-    _ -> do me <- liftIO . runReaderT (try $ discordOnStart opts) $ handle
-            case me of
-              Left (e :: SomeException) -> libError ("discordOnStart handler stopped on an exception:\n\n" <> T.pack (show e))
-              Right _ -> loop
+    Right (user, app) -> do putMVar (cacheHandleCache (discordHandleCache handle)) (Right (createCache user app))
+                            me <- liftIO . runReaderT (try $ discordOnStart opts) $ handle
+                            case me of
+                              Left (e :: SomeException) -> libError ("discordOnStart handler stopped on an exception:\n\n" <> T.pack (show e))
+                              Right _ -> loop
  where
+   startupRestCalls :: IO (Either RestCallInternalException (User, FullApplication))
+   startupRestCalls = do resp <- writeRestCall (discordHandleRestChan handle) R.GetCurrentUser
+                         case resp of
+                           Left e -> pure (Left e)
+                           Right user -> do
+                             resp2 <- writeRestCall (discordHandleRestChan handle) R.GetCurrentApplication
+                             case resp2 of
+                               Left e -> pure (Left e)
+                               Right app -> pure (Right (user, app))
+
    libError :: T.Text -> IO T.Text
    libError msg = tryPutMVar (discordHandleLibraryError handle) msg >> pure msg
 
--- a/src/Discord/Internal/Gateway/Cache.hs
+++ b/src/Discord/Internal/Gateway/Cache.hs
@@ -20,7 +20,7 @@ data Cache = Cache
      , cacheDMChannels :: !(M.Map ChannelId Channel)
      , cacheGuilds :: !(M.Map GuildId (Maybe (Guild, Maybe GuildCreateData)))
      , cacheChannels :: !(M.Map ChannelId Channel)
-     , cacheApplication :: !PartialApplication
+     , cacheApplication :: !FullApplication
      } deriving (Show)
 
 data CacheHandle = CacheHandle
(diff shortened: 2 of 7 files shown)


src/Discord.hs
```diff
@@ -116,18 +117,29 @@ runDiscord opts = do
...
  where
+   startupRestCalls :: IO (Either RestCallInternalException (User, FullApplication))
+   startupRestCalls = do resp <- writeRestCall (discordHandleRestChan handle) R.GetCurrentUser
+                         case resp of
+                           Left e -> pure (Left e)
+                           Right user -> do
+                             resp2 <- writeRestCall (discordHandleRestChan handle) R.GetCurrentApplication
+                             case resp2 of
+                               Left e -> pure (Left e)
+                               Right app -> pure (Right (user, app))
```
aquarial, line 141: Open to suggestions for a more haskell way to write this

src/Discord/Internal/Gateway/Cache.hs
```diff
@@ -20,25 +20,19 @@ data Cache = Cache
      , cacheDMChannels :: !(M.Map ChannelId Channel)
      , cacheGuilds :: !(M.Map GuildId (Maybe (Guild, Maybe GuildCreateData)))
      , cacheChannels :: !(M.Map ChannelId Channel)
-     , cacheApplication :: !PartialApplication
+     , cacheApplication :: !FullApplication
```
aquarial, line 23: The [Ready event](https://discord.com/developers/docs/topics/gateway-events#ready) contains a PartialApplication (just id and flags are set) wheras the [GetCurrentApplication](https://discord.com/developers/docs/resources/application) contains all the fields. 

Not sure what to name it exactly. I also considered `Application` (too common? overlap with user types?).

src/Discord/Internal/Types/ApplicationInfo.hs
```diff
@@ -0,0 +1,26 @@
...
+  , fullApplicationName :: T.Text
+  , fullApplicationFlags :: Int
+  } deriving (Show, Eq, Read)
+
+instance FromJSON FullApplication where
+  parseJSON = withObject "FullApplication" $ \o ->
+    FullApplication <$> o .: "id"
+                    <*> o .: "name"
+                    <*> o .: "flags"
+
```
aquarial, line 26: Can be filled in later with all the fields https://discord.com/developers/docs/resources/application

src/Discord.hs
L0neGamer:
```haskell
do
  eUser <- writeRestCall (discordHandleRestChan handle) R.GetCurrentUser
  eApp <- writeRestCall (discordHandleRestChan handle) R.GetCurrentApplication
  pure $ (,) <$> eUser <*> eApp
```
would be an alternative way of writing this I think.

Another alternative would be to match on `(Right user, Right app)` and then have two further cases for when either is Left, which would be equivalent to what you have but with less nesting.

L0neGamer:
I'll admit, initially I wasn't on board with making an additional request at the beginning here. But now that I see we already do that, and don't use it to boot, I'm happier with this.

I am still a bit iffy on us using an MVar for the cache as I slightly fear that it won't be properly filled, however I think in this case it should be fine? we might be able to rejig it but it'll add complexity I think, and I'm not sure it's worth it in this case.

There's a couple other refactors that could be done here but I can do them after we merge.

src/Discord/Internal/Gateway/Cache.hs
```diff
@@ -20,25 +20,19 @@ data Cache = Cache
...
-      ready <- readChan eventChan
-      case ready of
-        Right (InternalReady _ user _ _ _ _ pApp) -> do
-          putMVar cache (Right (Cache user M.empty M.empty M.empty pApp))
-          loop
-        Right r ->
-          writeChan log ("cache - stopping cache - expected Ready event, but got " <> T.pack (show r))
-        Left e ->
-          writeChan log ("cache - stopping cache - gateway exception " <> T.pack (show e))
+cacheLoop isEnabled cacheHandle _log = loop
```
L0neGamer, line 35: as my other comment said, a refactor could be that we do `when isEnabled loop`, which means that the cache thread won't just be spinning endlessly in the background. this would also simplify the `loop`.

src/Discord/Internal/Gateway/Cache.hs
```diff
@@ -20,25 +20,19 @@ data Cache = Cache
      , cacheDMChannels :: !(M.Map ChannelId Channel)
      , cacheGuilds :: !(M.Map GuildId (Maybe (Guild, Maybe GuildCreateData)))
      , cacheChannels :: !(M.Map ChannelId Channel)
-     , cacheApplication :: !PartialApplication
+     , cacheApplication :: !FullApplication
      } deriving (Show)
 
 data CacheHandle = CacheHandle
   { cacheHandleEvents :: Chan (Either GatewayException EventInternalParse)
   , cacheHandleCache  :: MVar (Either (Cache, GatewayException) Cache)
```
L0neGamer, line 28: Like I did in my branch I don't think there's anything useful about storing the gatewayexception here alongside the cache. In fact, in the `loop` if we ever get a gateway exception we refuse to update the cache, which is a mistake imo

src/Discord/Internal/Rest/ApplicationInfo.hs
```diff
@@ -0,0 +1,38 @@
...
+
+instance Request (FullApplicationRequest a) where
+  majorRoute =  applicationMajorRoute
+  jsonRequest = applicationJsonRequest
+
+
+-- | Data constructor for requests. See <https://discord.com/developers/docs/resources/ API>
+data FullApplicationRequest a where
+  -- | Get the full application for the current user
+  GetCurrentApplication :: FullApplicationRequest FullApplication
```
L0neGamer, line 27: can't believe we didn't have that, good call

L0neGamer:
The compilation failures are from not including the ApplicationInfo files in the cabal exposed-modules. Adding those makes it compile. I also ran the `cache` example and it works.

I think it would be worth mentioning in a comment somewhere what sort of data the cache will always have, and what sort of data it will only have if the cache is enabled (guild and channel info etc).

src/Discord/Internal/Gateway/Cache.hs
aquarial: This plus your `when isEnabled loop` suggestion greatly simplified `cacheLoop`

src/Discord/Internal/Types/ApplicationInfo.hs
L0neGamer: Could you make a ticket for this?

src/Discord.hs
```diff
@@ -116,18 +117,24 @@ runDiscord opts = do
...
 runDiscordLoop :: DiscordHandle -> RunDiscordOpts -> IO T.Text
 runDiscordLoop handle opts = do
-  resp <- liftIO $ writeRestCall (discordHandleRestChan handle) GetCurrentUser
+  resp <- startupRestCalls
   case resp of
     Left (RestCallInternalErrorCode c e1 e2) -> libError $
              "HTTP Error Code " <> T.pack (show c) <> " " <> TE.decodeUtf8 e1
                                                    <> " " <> TE.decodeUtf8 e2
     Left (RestCallInternalHttpException e) -> libError ("HTTP Exception -  " <> T.pack (show e))
     Left (RestCallInternalNoParse _ _) -> libError "Couldn't parse GetCurrentUser"
```
L0neGamer, line 126: this error is technically incomplete now, as the Left value could also be from current application. no big issue though

L0neGamer:
I'd be happy to approve this pending the ticket creation and a comment or two of documentation about when fields of the cache will be filled.

src/Discord.hs
aquarial: Generalized the error. 

Added issue for specific ADT https://github.com/discord-haskell/discord-haskell/issues/192

aquarial:
User relevant changes:
- The cache is now filled with user/application before the onStart handler
- The cache now contains FullApplication instead of PartialApplication. Minor breaking change for things like [the application command example](https://github.com/discord-haskell/discord-haskell/pull/190/commits/bb8328642a97cacc24b676bcf72b9f471ebb5bc4)

Fixes https://github.com/discord-haskell/discord-haskell/issues/186

aquarial:
I'll update the cache doc in a few hours, gtg

L0neGamer (approved):
Good enough. If you see something else you want to change I'd request mentioning the setting for enabling the cache around the cache data structure documentation, but it should be fine.

L0neGamer (approved):
lgtm!

PR #191, opened 2023-09-10
Fix Guild Create

L0neGamer:
A change I made was to make getting the guild id unfailable in cases where we are given a guild channel. however, when we receive a guild create event, the guild id may not be set.

this PR fixes that by creating a more intelligent parser that ensures that the guild id will be present when parsing the guild create data.


aquarial (approved):
Fun fix to inject the gid into the json before reparsing it.

Def needs to merge

PR #197, opened 2023-10-03
Change ReactionTiming type to match Discord docs

penelopeysm:
Hi 🙂  Couple of fixes to the `ReactionTiming` type

- It takes a `UserId`, not a `MessageId` (cf. https://discord.com/developers/docs/resources/channel#get-reactions).
- The endpoint doesn't take a `before` query parameter so I removed that.
- After some experimentation, I think that omitting the `after` query parameter doesn't actually give the latest batch of reactions. It gives the first batch of reactions, which are ordered by user ID -- so effectively it fetches the reactions of the oldest users. I changed the constructor names to reflect this.

### DIFF
--- a/src/Discord/Internal/Rest/Channel.hs
+++ b/src/Discord/Internal/Rest/Channel.hs
@@ -159,16 +159,14 @@ instance Default MessageDetailedOpts where
                             }
 
 -- | Data constructor for `GetReactions` requests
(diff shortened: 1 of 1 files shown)


L0neGamer:
I'll review this tonight, sorry about the delay

L0neGamer (approved):
lgtm!

PR #198, opened 2023-11-28
make Snowflake type derive num

aquarial:
Small change to allow int literals to be casted to Num.

Changes `DiscordId (Snowflake 11111)` to `DiscordId 111111`


aquarial:
Honestly a good enough reason against it that Snowflakes are not numbers. Shouldn't add,mutliply,etc.

PR #199, opened 2023-12-08
added: default for StartThreadOpt, THREAD_MEMBER_UPDATE; changed: comment for ThreadMembersUpdate

S3NP41-v:
default instance for StartThreadOpt
event for [ThreadMemberUpdate](https://discord.com/developers/docs/topics/gateway-events#thread-member-update)
changed the comment for ThreadMembersUpdate [according to the docs](https://discord.com/developers/docs/topics/gateway-events#thread-members-update)

### DIFF
--- a/src/Discord/Internal/Rest/Channel.hs
+++ b/src/Discord/Internal/Rest/Channel.hs
@@ -308,6 +308,12 @@ instance ToJSON StartThreadOpts where
       , "rate_limit_per_user" .=? startThreadRateLimit
       ]
 
+instance Default StartThreadOpts where
+  def = StartThreadOpts { startThreadName        = ""
+                        , startThreadAutoArchive = Nothing
+                        , startThreadRateLimit   = Nothing
+                        }
+
 -- | Options for `StartThreadNoMessage` request
 data StartThreadNoMessageOpts = StartThreadNoMessageOpts
   { -- | Base options for the thread
--- a/src/Discord/Internal/Rest/Channel.hs
+++ b/src/Discord/Internal/Rest/Channel.hs
@@ -309,7 +309,7 @@ instance ToJSON StartThreadOpts where
       ]
 
 instance Default StartThreadOpts where
-  def = StartThreadOpts { startThreadName        = ""
+  def = StartThreadOpts { startThreadName        = "New Thread"
                         , startThreadAutoArchive = Nothing
                         , startThreadRateLimit   = Nothing
                         }
--- a/src/Discord/Internal/Rest/Guild.hs
+++ b/src/Discord/Internal/Rest/Guild.hs
@@ -250,23 +250,6 @@ data CreateGuildChannelOpts
   | CreateGuildChannelOptsCategory
   deriving (Show, Read, Eq, Ord)
 
--- | result for `ListGuildActiveThreads`
-data ListActiveThreads = ListActiveThreads
-  { listActiveThreadsThreads :: [Channel]
-  , listActiveThreadsMembers :: [ThreadMember]
-  } deriving (Show, Read, Eq, Ord)
-
-instance ToJSON ListActiveThreads where
-  toJSON ListActiveThreads{..} = object
-    [ ("threads", toJSON listActiveThreadsThreads)
-    , ("members", toJSON listActiveThreadsMembers)
-    ]
-
-instance FromJSON ListActiveThreads where
-  parseJSON = withObject "ListActiveThreads" $ \o ->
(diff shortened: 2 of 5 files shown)


src/Discord/Internal/Types/Events.hs
```diff
@@ -41,11 +41,13 @@ data Event =
   | ThreadCreate               Channel
   -- | thread was updated
   | ThreadUpdate               Channel
+  -- | thread member for the current user was updated
+  | ThreadMemberUpdate         GuildId
```
L0neGamer, line 45: Counter to what I said, I'm pretty sure this event gives a thread member structure with a guild id as an additional field; would it be possible to give the thread member and the guild id? additionally, are you sure that this event parses? I feel like you don't parse out the `guild_id` sufficiently.

src/Discord/Internal/Rest/Channel.hs
```diff
@@ -308,6 +308,12 @@ instance ToJSON StartThreadOpts where
       , "rate_limit_per_user" .=? startThreadRateLimit
       ]
 
+instance Default StartThreadOpts where
+  def = StartThreadOpts { startThreadName        = ""
```
L0neGamer, line 312: reading the docs it looks like the thread name must be at least one character; maybe we can add that to the above documentation, as well as set a default thread name of `no thread name` or similar to make sure the default works?

S3NP41-v: maybe the name of `New Thread`? As discord suggests by default
obraz

src/Discord/Internal/Types/Events.hs
S3NP41-v: should i create a ThreadMemberUpdate record that is the thread member with a guild id, also being honest i am not very sure if it parses

L0neGamer: `gateway - received exception [Error in $: Error in $: parsing Snowflake failed, expected String, but encountered Object] while decoding {"t":"THREAD_MEMBER_UPDATE","s":13,"op":0,"d":{"user_id":"?","muted":false,"mute_config":null,"join_timestamp":"2023-12-08T18:40:00.424615+00:00","id":"?","guild_id":"?","flags":0}}` yeah it doesn't parse currently

L0neGamer:
Just want to check that a new feature works, and then I can approve!

src/Discord/Internal/Rest/Guild.hs
```diff
@@ -53,6 +53,8 @@ data GuildRequest a where
   --   'MANAGE_CHANNELS' permission. Returns a list of all of the guild's 'Channel'
   --   objects on success. Fires multiple Channel Update 'Event's.
   ModifyGuildChannelPositions      :: GuildId -> [(ChannelId,Int)] -> GuildRequest [Channel]
+  -- | Returns a list of active threads in a guild
+  ListActiveGuildThreads   :: GuildId -> GuildRequest ListActiveThreads
```
L0neGamer, line 57: Have you tested this works and parses what discord gives us?

S3NP41-v: i have tested it, but please do so also

S3NP41-v: yeah went back to look and noticed that i did not add an export for `ListActiveGuildThreads`
maybe `ListActiveThreads` should be moved to `Internal.Types.Guild` ?

S3NP41-v: > i have tested it, but please do so also

adding an image of a parsed return
obraz

S3NP41-v: > yeah went back to look and noticed that i did not add an export for `ListActiveThreads` maybe it should be moved to `Internal.Types.Guild` ?

i think that would be best so im going to do that

S3NP41-v: > yeah went back to look and noticed that i did not add an export for `ListActiveThreads` maybe it should be moved to `Internal.Types.Guild` ?

correction, not `Internal.Types.Guild` but `Internal.Types.Channel`

L0neGamer: I disagree with the movement; the List Active Threads structure is only ever used in the output of this one thing, so it should be defined and exported from this file as well; it's a request type.

S3NP41-v: reverted changes and added an export in `Internal.Rest.Guild`

L0neGamer (approved):
looks good!

PR #200, opened 2024-01-27
Add `Default` Instance for `InteractionResponseMessage`

julmb:
The type `InteractionResponseMessage` consists entirely of `Maybe` fields, so it works well with `Default`. This makes creating new response messages easier.

### DIFF
--- a/src/Discord/Internal/Types/Embed.hs
+++ b/src/Discord/Internal/Types/Embed.hs
@@ -56,6 +56,8 @@ createEmbed CreateEmbed{..} =
            , embedProvider    = Nothing
            }
 
+-- | The 'Default' instance of this type yields a 'def' value whose fields are all empty.
+-- As such, the 'def' value is not a valid embed and needs to be adjusted before being used.
 data CreateEmbed = CreateEmbed
   { createEmbedAuthorName  :: T.Text
   , createEmbedAuthorUrl   :: T.Text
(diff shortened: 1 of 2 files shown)


L0neGamer:
I believe the reason I didn't do this originally is that the data type requires at least one of the fields to be filled in, meaning that a Default implementation would break by default, hence the basic constructor which fills in the message field defined above.

julmb:
Ah, I see, I did not realize that, that makes sense. However, it is a little unfortunate since the basic helper function forces it to have a content field, even though a response message consisting entirely of embeds is also valid. I guess one way to deal with this would be to make more helper functions, but not sure if that is a good idea.

julmb:
Also, there seems to be a precedent for this: `CreateEmbed` has a `Default` instance whose `def` value leads to invalid requests when used directly in a message. Is there a guideline that says that `def` needs (or does not need) to be a valid value when used directly?

L0neGamer:
Of course, you can overwrite the other fields freely, so it's mostly a question of what we assume the default implementation should be; by default, will a user just send a text response, or by default will they do one of the other constructions?

If you add a haddock comment to the data type detailing this behaviour then I think the Default instance will be fine.

L0neGamer:
Good point on CreateEmbed; could I trouble you to similarly comment that data type?

julmb:
I added notes to the documentation of both `InteractionResponseMessage` and `CreateEmbed`.

L0neGamer (approved):
lgtm!

PR #204, opened 2024-02-29
Add ToJSONKey and FromJSONKey instances for Snowflake and DiscordId

julmb:
This could be useful for people who want to have something like `Map GuildId Config` to store a separate configuration per guild.

### DIFF
--- a/src/Discord/Internal/Types/Prelude.hs
+++ b/src/Discord/Internal/Types/Prelude.hs
@@ -108,6 +108,12 @@ instance FromJSON Snowflake where
             (Just i) -> pure i
       )
 
+instance ToJSONKey Snowflake where
+  toJSONKey = contramapToJSONKeyFunction unSnowflake toJSONKey
+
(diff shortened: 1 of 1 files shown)


L0neGamer (approved):
I was worried we were going to run into the object Vs array issues of jsonkey, but word64 has a sensible implementation so that is fine.

julmb:
Thank you for the quick merge!

I just realized that `ToJSONKeyFunction` has a `Contravariant` instance, so we could use `contramap` or `>$<` instead of the unwieldly `contramapToJSONKeyFunction` function. However, it would require an extra import, and I'm not sure it is worth messing with this any further.

L0neGamer:
I'm working on an MR that would remove the need for these instances; I'll link it here

L0neGamer:
https://github.com/discord-haskell/discord-haskell/pull/205 like so

L0neGamer:
Nw, without your PR I wouldn't have thought to do this.

PR #206, opened 2024-03-24
Implemented Audit log and auto moderation rules

S3NP41-v:
there were a few places where i am not so sure what convention to take upon, namely where to put `toAuditLogEvent` from `Discord\Internal\Rest\Guild`, and also i was forced to take liberties with the discord api doc's, as they don't list a few fields from auto moderation rules as optional, but in practice i found out that, they are often omitted by discord which caused exceptions with the parser that expected those fields.

### DIFF
--- a/src/Discord/Internal/Rest/Guild.hs
+++ b/src/Discord/Internal/Rest/Guild.hs
@@ -16,6 +16,8 @@ module Discord.Internal.Rest.Guild
   , CreateGuildIntegrationOpts(..)
   , ModifyGuildIntegrationOpts(..)
   , ListActiveThreads (..)
+  , GetAuditLogOpts (..)
+  , toAuditLogEvent
   ) where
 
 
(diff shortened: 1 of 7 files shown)


src/Discord/Internal/Rest/Guild.hs
```diff
@@ -320,6 +329,137 @@ instance FromJSON ListActiveThreads where
...
+                   <*> o .:? "old_value"
+
+
+-- | This is a temporary solution, i really hope.
+--   Discord did NOT think this through, read more at <https://discord.com/developers/docs/resources/audit-log#audit-log-change-object>
+data AuditLogChangeValue = AuditLogChangeValue String
+  deriving ( Show, Read )
+
+instance FromJSON AuditLogChangeValue where
+  parseJSON o = return $ AuditLogChangeValue (show o)
```
L0neGamer, line 404: ah I see what the issue is with the change object. Because just about anything can be marked as changed or whatever, just about any object can be included here. I think it would be for the best that the `Value` is stored instead of auditlogchangevalue, since the user can at least use aeson operations on it themselves if they wish.

L0neGamer (commented):
one minor change request!

src/Discord/Internal/Rest/Guild.hs
S3NP41-v: That is definitely a better solution!
Sad that discord put no thought behind this though...
I'll change that shortly

L0neGamer (approved):

PR #207, opened 2024-03-29
gateway events for audit log and auto moderation

S3NP41-v:
added a new file for audit log types, moved the audit log related types from `Internal.Rest.Guild.hs` to `Internal.Types.AuditLog.hs` and wrote gateway events for audit log and auto moderation rules.


S3NP41-v:
its fine, i just use my local version until approved, then i go back to the default, i also don't have all the free time so i understand.

L0neGamer (approved):

PR #211, opened 2024-10-29
Interaction fixes

L0neGamer:


yutotakano (approved):
Approving since it looks OK. I like the backwards-compatible pattern synonym. But the rename of SelectMenuData etc (for avoiding conflict with the other similarly named identifier from `Internal.Types.Components`) isn't very clean -- eventually we ought to fix this with some API changes.

PR #218, opened 2025-02-23
Support GHC 9.10

yutotakano:
Added the new base version as supported, and a tested-with clause for GHC 9.10 for Hackage visibility and the CI matrix.

Although that was enough to build on GHC 9.10, for future-proofness I've also bumped the upper bounds for dependencies where possible since the changelogs for the bumps didn't affect us.


L0neGamer (approved):
Lgtm! Merge when ready

Issue #232, opened 2026-01-03
Editing a post and removing embedded images


dixonary:
Doing the following:

1. Create a message using `CreateMessageDetailed` that includes one or more embeds, each with an embedded image
2. Edit that message to remove the embeds

Causes the message to still contain the embedded images.

On the other hand, changing the embedded image URL to a different embedded image URL seems to remove it forever! It flickers to the new image and then is removed about half a second later. Refreshing the client does not bring it back.

These two behaviours might be related or they might not. I can write a reproducer if helpful.

dixonary:
If the embeds were thumbnails, they become "full size" embedded images if the embeds are edited away.

L0neGamer:
Reproducer is always helpful.
I'm slightly worried that this is just expected discord behaviour, but it's something we can look into.

dixonary:
As is traditional I'm having trouble reproducing the first problem in a clean room.

Here's a reproducer for the second problem ("Changing the embedded image URL removes it forever"), which I think is more likely of the two to be a D-H bug:

```hs
{-# LANGUAGE LambdaCase, OverloadedStrings #-}

-- Only need `base >= 4.18` and `discord-haskell`
import Control.Concurrent
import Control.Monad (void)
import Control.Monad.IO.Class
import Discord
import Discord.Requests
import Discord.Types

main :: IO ()
main = do
  let
    message1 = def{messageDetailedEmbeds = Just [def{createEmbedThumbnail = Just (CreateEmbedImageUrl "https://dixonary.co.uk/static/purple-square.png")}]}
    message2 = def{messageDetailedEmbeds = Just [def{createEmbedThumbnail = Just (CreateEmbedImageUrl "https://dixonary.co.uk/static/green-square.png")}]}
  void $
    runDiscord
      def
        { discordToken = "TOKEN_HERE"
        , discordOnEvent = \case
            MessageCreate m
              | userIsBot (messageAuthor m) -> pure ()
              | messageContent m == "SHOW ME" -> do
                  r <- restCall (CreateMessageDetailed (messageChannelId m) message1)
                  case r of
                    Left _ -> pure ()
                    Right m' -> do
                      liftIO $ putStrLn "Wait 5..." >> threadDelay 5000000
                      void $ restCall $ EditMessage (messageChannelId m', messageId m') message2
            _ -> pure ()
        }
```

Invite it in and type "SHOW ME".

dixonary:
And here's one for the original problem:

```hs
[...]

PR #233, opened 2026-01-03
Implement Attachments more fully

L0neGamer:
In this PR we properly implement some attachments stuff, which lets users properly set, ignore, and overwrite attachments. We handle the encoding for them in this case.

Probably needs some cleanup which I can get to in time, but would appreciate an eye for oversight.

#232


L0neGamer:
Will be closed by #246

L0neGamer:
closed by #246

PR #235, opened 2026-04-01
Basic implementation of Polls

S3NP41-v:
to be entirely honest i would like to request additional time to work on this a bit more, as i got tired nearing the end and kind of rushed it,

I did check that `GetAnswerVoters` and `EndPoll` requests work, but the check was surface level and could not have cleared all possible edge cases,

Additionally i want to rethink the implementation of a few structures.


L0neGamer:
No rush on getting this ready; I'll mark it as a draft, and you can mark it as ready when you think it's suitable for review.

Would be nice to see links to relevant discord docs in the documentation if you can!

PR #237, opened 2026-04-30
Unwrap CreateEmbed Parameter

julmb:
This function is only ever called with the parameter wrapped in `Just` at the call site, so the `Maybe` can be removed.


L0neGamer (approved):
This makes it more composable as well!

julmb:
Closing as the code this touches no longer exists after #246.

PR #240, opened 2026-05-09
Enable Using Attachments in Interaction Responses

julmb:
Preliminary [discussion on this topic](https://discord.com/channels/918577626954739722/918577626954739726/1198507734136528897) happened in early 2024.

This PR adds the ability to create attachments when sending interaction responses.
- This is based on #237 and #239. I will rebase this PR to apply cleanly once those are merged.
- The first three commits (e6238831af2d2f8f6d506bf59c51da8e8bf39bae 6623aeb510fa2b7a8a5d4d6a5c01184736a3526b dd88adf8124d578ce6923b277f998578a0727271) are pure refactorings without any change in functionality. I could extract this into a separate PR if wanted.
- Adds `CreateAttachment`. `CreateAttachment` relates to `Attachment` like `CreateEmbed` relates to `Embed`.
- The field `createAttachmentId` is of type `Int` instead of `AttachmentId` since these seem to be indices rather than proper snowflake values. The snowflake values will only be created when the attachment is actually uploaded.
- In `InteractionResponseMessage`, in the type of `interactionResponseMessageAttachments`, changes `Attachment` to `CreateAttachment`.

I feel like the previous `Attachment` value in `InteractionResponseMessage` could not be used in a meaningful way, since it requires an attachment to already exist, and as far as I can tell, there was no way to create one until now. So this change should not restrict any uses that were previously possible, but I am not 100% sure.

I briefly tested the following cases to work correctly:
- Creating a response with an attachment
- Replacing a deferred response with a response containing an attachment
- Adding an attachment to an existing response
- Adding a second attachment to an existing response
- Removing an attachment from an existing response
- Replacing an attachment in an existing response

It might make sense to also use `CreateAttachment` in `MessageDetailedOpts`, which currently only allows attaching a single file. I will look into this when I get the chance.

### DIFF
--- a/src/Discord/Internal/Types/Channel.hs
+++ b/src/Discord/Internal/Types/Channel.hs
@@ -15,6 +15,8 @@ module Discord.Internal.Types.Channel (
   , Message (..)
   , AllowedMentions (..)
   , MessageReaction (..)
+  , CreateAttachment (..)
+  , attachmentPart
   , Attachment (..)
   , Nonce (..)
   , MessageReference (..)
@@ -32,11 +34,15 @@ import Control.Applicative (empty)
 import Data.Aeson
 import Data.Aeson.Types (Parser)
 import Data.Default (Default, def)
+import Data.ByteString (ByteString)
 import Data.Text (Text)
 import Data.Time.Clock
 import qualified Data.Text as T
 import Data.Bits
 import Data.Data (Data)
+import Text.Printf
+import Network.HTTP.Client
+import Network.HTTP.Client.MultipartFormData
 
 import Discord.Internal.Types.Prelude
 import Discord.Internal.Types.User (User(..), GuildMember)
@@ -668,6 +674,24 @@ instance ToJSON MessageReaction where
       , "me" .== messageReactionMeIncluded
       , "emoji" .== messageReactionEmoji
       ]
+
+data CreateAttachment = CreateAttachment
+  { createAttachmentId :: Int
+  , createAttachmentFilename :: Text
+  , createAttachmentData :: ByteString
+  } deriving (Show, Read, Eq, Ord)
+
+instance ToJSON CreateAttachment where
+  toJSON CreateAttachment {..} = objectFromMaybes
+    [ "id" .== createAttachmentId
+    , "filename" .== createAttachmentFilename
+    ]
+
+attachmentPart :: CreateAttachment -> PartM IO
+attachmentPart CreateAttachment {..} = partFileRequestBody name path body where
+  name = T.pack $ printf "files[%d]" createAttachmentId
+  path = T.unpack createAttachmentFilename
+  body = RequestBodyBS createAttachmentData
 
 -- | Represents an attached to a message file.
 data Attachment = Attachment
--- a/src/Discord/Internal/Rest/Channel.hs
+++ b/src/Discord/Internal/Rest/Channel.hs
@@ -518,7 +518,7 @@ channelJsonRequest c = case c of
                     (RequestBodyBS $ snd f)
                 ]
           )
-            ++ join (maybe [] (maybeEmbed . Just <$>) (messageDetailedEmbeds msgOpts))
+            ++ join (maybe [] (embedPart <$>) (messageDetailedEmbeds msgOpts))
 
         payloadData =  objectFromMaybes $
                         [ "content" .== messageDetailedContent msgOpts
@@ -573,7 +573,7 @@ channelJsonRequest c = case c of
                     (RequestBodyBS $ snd f)
                 ]
           )
-            ++ join (maybe [] (maybeEmbed . Just <$>) (messageDetailedEmbeds msgOpts))
+            ++ join (maybe [] (embedPart <$>) (messageDetailedEmbeds msgOpts))
 
         payloadData =  objectFromMaybes $
                         [ "content" .== messageDetailedContent msgOpts
--- a/src/Discord/Internal/Rest/Interactions.hs
+++ b/src/Discord/Internal/Rest/Interactions.hs
@@ -79,11 +79,13 @@ interactionResponseJsonRequest a = case a of
     Delete (interaction aid it /~ mid) mempty
   where
     bodyIR :: InteractionResponse -> RestIO R.ReqBodyMultipart
-    bodyIR ir@(InteractionResponseChannelMessage irm) = R.reqBodyMultipart (partBS "payload_json" (BL.toStrict $ encode ir) : files irm)
-    bodyIR ir@(InteractionResponseUpdateMessage irm) = R.reqBodyMultipart (partBS "payload_json" (BL.toStrict $ encode ir) : files irm)
-    bodyIR ir = R.reqBodyMultipart [partBS "payload_json" $ BL.toStrict $ encode ir]
+    bodyIR ir@(InteractionResponseChannelMessage irm) = R.reqBodyMultipart $ payload ir : files irm
+    bodyIR ir@(InteractionResponseUpdateMessage irm) = R.reqBodyMultipart $ payload ir : files irm
+    bodyIR ir = R.reqBodyMultipart [payload ir]
     bodyIRM :: InteractionResponseMessage -> RestIO R.ReqBodyMultipart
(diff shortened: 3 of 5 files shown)


src/Discord/Internal/Types/Channel.hs
```diff
@@ -669,6 +675,24 @@ instance ToJSON MessageReaction where
...
+  { createAttachmentId :: Int
+  , createAttachmentFilename :: Text
+  , createAttachmentData :: ByteString
+  } deriving (Show, Read, Eq, Ord)
+
+instance ToJSON CreateAttachment where
+  toJSON CreateAttachment {..} = objectFromMaybes
+    [ "id" .== createAttachmentId
+    , "filename" .== createAttachmentFilename
+    ]
```
L0neGamer, line 688: is this ToJSON instance useful?

L0neGamer (commented):
I'm not the most confident with attachments stuff, but I did recently make a PR for them in https://github.com/discord-haskell/discord-haskell/pull/233; would your work here be covered by this?

src/Discord/Internal/Types/Channel.hs
L0neGamer: I'm not sure I enjoy dropping a field from a data type.

julmb:
> I'm not the most confident with attachments stuff, but I did recently make a PR for them in #233; would your work here be covered by this?

I completely missed this PR, sorry. :(

The two PRs are conceptually orthogonal, but difficult to merge. In particular, #233 completely removes the functionality that this PR is enabling (interaction response attachments). I was not aware that embed attachments were broken/incomplete in `master` (missing `attachments` key in `payload_json`), but it now makes sense that they would clash.

If both regular message attachments and embed attachments are supposed to work correctly, these two facilities need to "collaborate", since they both add entries to the single `attachments` list in the request. This is further complicated if we want to give users control over these attachments (letting them add/remove/replace both message and embed attachments).

I will have to think about this a little more.

src/Discord/Internal/Types/Channel.hs
julmb: Values of type `CreateAttachment` are transmitted to the Discord API in two parts:
- The fields `createAttachmentId` and `createAttachmentFilename` are transmitted in the `payload_json` part of the multipart request, hence the `ToJSON` instance
- The field `createAttachmentData` is transmitted directly as a separate part of the multipart request, hence the `attachmentPart` function

I think the question is whether we want to represent values that are conceptually complete (an attachment is conceptually a single value that has a name and some data) or values that correspond exactly to partial pieces of information that are sent to the Discord API (an attachment has a metadata part and a data part, the first is sent via JSON and the second one via a multipart part, and those are separate Haskell types).

julmb:
Okay, some more thoughts.

The way things are currently designed, a `CreateEmbed` value can contain up to 4 `ByteString` values for images that get turned into attachments when that `CreateEmbed` appears in a message. This hides the underlying mechanism of turning the metadata into JSON values and the data into multipart parts, and connecting these two via numerical IDs in the form of `files[n]` identifiers.

This makes things very convenient for the user in simple use cases, but has a number of drawbacks:
- Complex processes cannot be represented. For example, creating an embed with a large image upload and then later changing parts of the message without reuploading the large image is impossible with the current design as far as I can tell.
- Special care has to be taken to generate both unambiguous indices and names. Currently, `createEmbedTitle` is prepended to each embed image. This process fails if there are multiple embeds with the same title, which as far as I can tell is allowed by Discord. I tested this and it is broken on both `master` and #233, albeit in different ways.
- Making this process work for message attachments and embed attachments is tricky. If we want to keep the approach that is currently used for embed attachments, then this will force message attachments to use something similar, as we cannot give the user control over some indices but not others.

The approach in this PR is kind of an awkward middle ground where users are burdened with managing attachment indices (the `files[n]` part) but they cannot actually control which entries appear in the JSON metadata (the `attachments` key) and which appear in the multipart parts to actually be uploaded. Although I haven't tested this against the Discord API. It sounds like they allow repeating `attachments` entries without reupload, but maybe it doesn't actually work.

I lean towards giving the users more flexibility, even if it means that they will have to interact with more technical details like managing attachment data and metadata separately. Maybe we can provide some convenience functions to make the common use case easier.

Caveat: I only started working on attachments a few days ago, so it is possible that I am misunderstanding something here. Maybe some of the maintainers can weigh in on which direction would best align with the design of `discord-haskell`.

L0neGamer:
My preference for discord-haskell is to give as much freedom to the user as possible, trying to avoid footguns if that is possible but ultimately letting them do what they wish. The fact we didn't (and still don't!) have attachments previously is kinda embarrassing, but we will have to push forwards.

Saying that, I do feel like having unique indices and unique filenames feels like an edge we _can_ sand off. As mentioned, #233 automatically generates filenames and indices; as I understand it the filename uniqueness issue could be solved by including the index in the filename, meaning that each list entry must have a different name. An alternative could be generating a hash based on the file contents, and so on.

Unfortunately as can be seen to 233 only recently being put up, this is not an area of the API I understand well. I wouldn't want to put another half measure in (hence my stab at a full implmentation with #233), but it's a question of how much rope we give users.

Ideally there'd be a wrapper library that makes discord-haskell nice to work with, but for now all we have is us.

julmb:
Okay, I just did some more tests on editing messages with attachments. It turns out that you can, for example, take a message with an existing attachment and add another attachment without reuploading the first attachment, but it is quite involved:
1. Create message with an attachment (`attachments` key in `payload_json` has one entry, upload the attachment via multipart form entry).
2. Retrieve the message that was just created (some d-h calls like `CreateInteractionResponse` do not return the message, although maybe that could be changed)
3. Extract the attachment snowflakes that discord created from the message data
4. Edit the message by supplying a new attachment record that contains snowflakes of the existing attachments and indices of the new attachments. Only upload the new attachments via multipart form entry.

I am not sure if it is possible to perform this process on the user's behalf without exposing them to some of the bookkeeping. If all the user supplies is some `ByteString` values, we would have to (statefully) keep track of which data was already submitted to Discord in order to avoid reuploading it. And even if we do that, it would lead to quite complicated behind-the-scenes behavior that the could surprise users.

We could of course decide that even though the Discord API supports it, d-h will not offer maximum flexibility here and that instead, all attachments are always reuploaded fully everytime a message is edited.

L0neGamer:
I think my preference is to allow a determined user to upload additional attachments without reuploading previous ones, but for the default to be "overwrite existing data" (requiring full reupload) if that makes sense?

Ideally I want someone to be able to use discord-haskell to write their own arbitrary discord api, but to do that we need to offer the full slew of capability.

I believe that CreateInteractionResponse from discord does not return the response by default: https://docs.discord.com/developers/interactions/receiving-and-responding#create-interaction-response. I think it would make sense for us to default getting a response to on, though, since it's pretty easy for a haskeller to discard received data.

Ultimately, for now I don't have time to adjust our existing solutions much; if you have time to do so over the next few days I'd be interested to see your implementation, happy to chat here or in discord.

julmb:
> I think my preference is to allow a determined user to upload additional attachments without reuploading previous ones, but for the default to be "overwrite existing data" (requiring full reupload) if that makes sense?
> 
> Ideally I want someone to be able to use discord-haskell to write their own arbitrary discord api, but to do that we need to offer the full slew of capability.

I agree with this approach. I think in order to enable the full capability, we need to give users fairly direct access to the Discord API. I think it should be possible to provide some convenience functionality to support the common use case better.

> I believe that CreateInteractionResponse from discord does not return the response by default: https://docs.discord.com/developers/interactions/receiving-and-responding#create-interaction-response. I think it would make sense for us to default getting a response to on, though, since it's pretty easy for a haskeller to discard received data.

Agreed, unless enabling this would come with a noticeable performance cost, but I wouldn't imagine so.

> Ultimately, for now I don't have time to adjust our existing solutions much; if you have time to do so over the next few days I'd be interested to see your implementation, happy to chat here or in discord.

I am travelling tomorrow and then the work week will strike back, so I can only work on this here and there. But I will try to put some more thought into it and if I come up with anything, I will post an update.

L0neGamer:
No rush at all; keep me posted, and thanks for the conversation.

julmb:
Closing this as superseded by #246.

PR #246, opened 2026-05-16
Fully-General Attachments and Uploads

julmb:
Guiding principle: A motivated user should be able to do anything the Discord API allows.

I considered the following example scenarios to guide the design:
1. Adding/removing some attachments to/from a message while keeping existing attachments intact without reuploading them.
2. Using the same attachment in multiple places (for example, author icon of multiple embeds) within the same message.

So far, this draft enables:
- Adding an arbitrary number of attachments to regular messages (used to be restricted to a single attachment)
- Adding an arbitrary number of attachments to interaction responses (used to be impossible)
- Adding and removing attachments from existing messages without reuploading existing files
- Fixing #232

It is worth noting that it achieves this without burdening the user with managing Discord's numeric file indices (`files[n]`). When uploading an attachment, the user only needs to provide a filename and the data.

This PR supersedes #240.

# `CreateEmbed`

The question that remains is how to deal with the existing `CreateEmbed` infrastructure. High-level pseudocode:
```hs
data CreateEmbed = CreateEmbed {
    title :: Text,
    image :: ByteString }

data CreateUpload = CreateUpload {
    filename :: Text,
    content :: ByteString }

data Message = Message {
    content :: Text,
    embeds :: [CreateEmbed],
    uploads :: [CreateUpload] }
```
`CreateEmbed` allows adding images to embeds by specifying them at the embed level directly as values. This makes (2) impossible, since there is no way of referencing attachments, which fundamentally exist at the level of messages, not at the level of individual embeds.

With the preliminary work in this draft, it would be very easy to replace the `image :: ByteString` with `image :: Text` and ask the user to reference their attachments via Discord's `attachment://upload.png` syntax. The user would then be responsible for managing the attachments at the message level and referencing them in each embed, giving them access to the full flexibility that the API exposes. This would also fix the disambiguation bug that currently exists when multiple embeds with the same title are added to the same message. It is even possible that this would make the entire `CreateEmbed` type obsolete, since it would then be nearly identical to `Embed` itself.

This would be my preferred way of doing things. However, it would be a breaking change for existing users of the `CreateEmbed` functionality. I have not spent too much time thinking about ways of preserving the `CreateEmbed` functionality, but I think it will be quite difficult to do so while also giving access to the new functionality in a consistent way.

### DIFF
--- a/src/Discord/Internal/Rest/Channel.hs
+++ b/src/Discord/Internal/Rest/Channel.hs
@@ -26,10 +26,8 @@ import Data.Aeson
 import Data.Default (Default, def)
 import Text.Emoji (emojiFromAlias)
 import qualified Data.Text as T
-import qualified Data.ByteString as B
 import qualified Data.ByteString.Lazy as BL
-import Network.HTTP.Client (RequestBody (RequestBodyBS))
-import Network.HTTP.Client.MultipartFormData (partFileRequestBody, partBS)
+import Network.HTTP.Client.MultipartFormData (partBS)
 import Network.HTTP.Req ((/:), (/~))
 import qualified Network.HTTP.Req as R
 
@@ -139,8 +137,10 @@ data MessageDetailedOpts = MessageDetailedOpts
     messageDetailedTTS                      :: Bool
   , -- | embedded rich content (up to 6000 characters)
     messageDetailedEmbeds                   :: Maybe [CreateEmbed]
-  , -- | the contents of the file being sent
-    messageDetailedFile                     :: Maybe (T.Text, B.ByteString)
+  , -- | metadata of existing attachments
+    messageDetailedAttachments              :: Maybe [CreateAttachment]
+  , -- | new uploads to attach to the message
+    messageDetailedUpload                   :: Maybe [CreateUpload]
   , -- | allowed mentions for the message
     messageDetailedAllowedMentions          :: Maybe AllowedMentions
   , -- | If `Just`, reply to the message referenced
@@ -155,7 +155,8 @@ instance Default MessageDetailedOpts where
   def = MessageDetailedOpts { messageDetailedContent         = ""
                             , messageDetailedTTS             = False
                             , messageDetailedEmbeds          = Nothing
-                            , messageDetailedFile            = Nothing
+                            , messageDetailedAttachments     = Nothing
+                            , messageDetailedUpload          = Nothing
                             , messageDetailedAllowedMentions = Nothing
                             , messageDetailedReference       = Nothing
                             , messageDetailedComponents      = Nothing
@@ -507,23 +508,16 @@ channelJsonRequest c = case c of
       in Post (channels /~ chan /: "messages") body mempty
 
   (CreateMessageDetailed chan msgOpts) ->
-    let fileUpload = messageDetailedFile msgOpts
-        filePart =
-          ( case fileUpload of
-              Nothing -> []
-              Just f ->
-                [ partFileRequestBody
-                    "file"
-                    (T.unpack $ fst f)
-                    (RequestBodyBS $ snd f)
-                ]
-          )
+    let fileAttachments = messageDetailedAttachments msgOpts
+        fileUpload = messageDetailedUpload msgOpts
+        filePart = maybe [] uploadsParts fileUpload
             ++ join (maybe [] (maybeEmbed . Just <$>) (messageDetailedEmbeds msgOpts))
 
         payloadData =  objectFromMaybes $
                         [ "content" .== messageDetailedContent msgOpts
                         , "tts"     .== messageDetailedTTS msgOpts ] ++
                         [ "embeds" .=? ((createEmbed <$>) <$> messageDetailedEmbeds msgOpts)
+                        , "attachments" .=? uploadsAdd fileUpload fileAttachments
                         , "allowed_mentions" .=? messageDetailedAllowedMentions msgOpts
                         , "message_reference" .=? messageDetailedReference msgOpts
                         , "components" .=? messageDetailedComponents msgOpts
@@ -562,23 +556,16 @@ channelJsonRequest c = case c of
 
   -- copied from CreateMessageDetailed, should be outsourced to function probably
   (EditMessage (chan, msg) msgOpts) ->
-    let fileUpload = messageDetailedFile msgOpts
-        filePart =
-          ( case fileUpload of
-              Nothing -> []
-              Just f ->
-                [ partFileRequestBody
-                    "file"
-                    (T.unpack $ fst f)
-                    (RequestBodyBS $ snd f)
-                ]
-          )
+    let fileAttachments = messageDetailedAttachments msgOpts
+        fileUpload = messageDetailedUpload msgOpts
+        filePart = maybe [] uploadsParts fileUpload
             ++ join (maybe [] (maybeEmbed . Just <$>) (messageDetailedEmbeds msgOpts))
 
         payloadData =  objectFromMaybes $
                         [ "content" .== messageDetailedContent msgOpts
                         , "tts"     .== messageDetailedTTS msgOpts ] ++
                         [ "embeds" .=? ((createEmbed <$>) <$> messageDetailedEmbeds msgOpts)
+                        , "attachments" .=? uploadsAdd fileUpload fileAttachments
                         , "allowed_mentions" .=? messageDetailedAllowedMentions msgOpts
                         , "message_reference" .=? messageDetailedReference msgOpts
                         , "components" .=? messageDetailedComponents msgOpts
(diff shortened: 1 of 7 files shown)


julmb:
Commit c2e2588 integrates the existing `CreateEmbed` functionality into the new upload infrastructure:
- Retains all the existing `CreateEmbed` functionality (including URL shape and associated disambiguation issues)
- Fixes #232
- Enables full flexibility of referenced attachments via `CreateEmbedImageUrl`

It works, but I would still prefer to not have this functionality at all, since it has some very surprising semantics. If you add an embed image to a message with existing attachments, that embed image will lead to an uploads entry, which will lead to the creation of an `attachments` key for the edit request. Unless the user also adds all of the existing attachments to the edit request, the existing attachments will disappear, even though to the user it looks like they never touched the `attachments` key and so they might expect these to be retained unchanged. There is no real way around this, as allowing the user to edit attachments (as required by #232) involves exposing them to the fact that embed images are also attachments, even though Discord tries its best to hide that fact from users. The "`ByteString` inside `CreateEmbed`" constellation is a leaky abstraction.

The surprising creation of an `attachments` key also happens with non-embed uploads, but I would argue that in this case, it is more obvious to the user that they have to specify which of the existing attachments should be retained in addition to the newly uploaded files.

I also found out yesterday that Discord will actually remove attachments from the `attachments` key if they are also used in embeds (thus also depriving the user from accessing the attachment's `id` and `filename`). This makes for some very surprising behavior which is further obfuscated by the magic that `CreateEmbed` does. Overall, `CreateEmbed` works in the current state of this PR, and it can simplify some use cases, but it comes with some surprising behavior that will be very difficult to understand since the user is so far removed from the actual calls that go to the API.

julmb:
I am now happy with the state of this PR. It fixes existing issues, enables new functionality, and retains all existing functionality with minimal breaking changes.

I am still wondering if it makes sense to remove the `CreateEmbedImageUpload` functionality entirely (and with it, potentially `CreateEmbed` itself). However, that can be done in a separate PR since this one now integrates all the existing infrastructure.

src/Discord/Internal/Types/Channel.hs
```diff
@@ -701,6 +711,58 @@ instance ToJSON Attachment where
       , "width" .=? attachmentWidth
       ]
 
+data CreateAttachment = CreateAttachment
```
L0neGamer: All of the additions in this module need haddocks/documentation on what they do and why they're needed; they may also do well to have links to the discord documentation.

src/Discord/Internal/Rest/Channel.hs
```diff
@@ -139,8 +136,10 @@ data MessageDetailedOpts = MessageDetailedOpts
     messageDetailedTTS                      :: Bool
   , -- | embedded rich content (up to 6000 characters)
     messageDetailedEmbeds                   :: Maybe [CreateEmbed]
-  , -- | the contents of the file being sent
-    messageDetailedFile                     :: Maybe (T.Text, B.ByteString)
+  , -- | metadata of existing attachments
+    messageDetailedAttachments              :: Maybe [CreateAttachment]
```
L0neGamer: if it's metadata of existing attachments, maybe `CreateAttachment` isn't the best name for this?

Maybe `SimpleAttachment`?

L0neGamer (changes requested):
I think this is a good direction to go in. I find myself a little confused about some of the difference between attachment and upload here.

I would appreciate an addition to our example files where the attachments and uploads of various files and messages are manipulated, both for general demonstration purposes and for my own observation purposes.

Blocker is more documentation.

src/Discord/Internal/Types/Channel.hs
```diff
@@ -701,6 +711,58 @@ instance ToJSON Attachment where
...
+  } deriving (Show, Read, Eq, Ord)
+
+uploadsEmbed :: CreateEmbed -> [CreateUpload]
+uploadsEmbed CreateEmbed {..} = catMaybes [author, thumbnail, image, footer] where
+  author = createEmbedAuthorIcon >>= go "author.png"
+  thumbnail = createEmbedThumbnail >>= go "thumbnail.png"
+  image = createEmbedImage >>= go "image.png"
+  footer = createEmbedFooterIcon >>= go "footer.png"
+  go _ (CreateEmbedImageUrl _) = Nothing
+  go name (CreateEmbedImageUpload dat) = Just $ CreateUpload (prefix <> name) Nothing Nothing dat
```
L0neGamer: another comment: do we want to have some kind of uniqueness string for the file names here? I feel like this has the same issues we talked about, where the prefix of the embed title may not be unique enough.

This could be somewhat solved by using the index of the embed in the uploads list or similar. 

I'd think we could have something like
```hs
uploadsEmbed :: Int -> CreateEmbed -> [CreateUpload]
uploadsEmbed uploadIndex CreateEmbed {..} = ...
  where
  ...
  prefix = T.show uploadIndex <> "-" <> T.filter (/= ' ') createEmbedTitle

uploadsAssemble :: Maybe [CreateEmbed] -> [CreateUpload] -> [CreateUpload]
uploadsAssemble embeds uploads = uploads ++ maybe [] (concat . zipWith uploadsEmbed [length uploads ..]) embeds
```

src/Discord/Internal/Rest/Channel.hs
julmb: The [discord documentation](https://docs.discord.com/developers/resources/message#create-message) calls this a "partial attachment object" but never defines what that means exactly. The [discord openapi spec](https://github.com/discord/discord-api-spec/) calls it a `MessageAttachmentRequest`.

I named it `CreateAttachment` in analogy to the `CreateEmbed`/`Embed` and `CreateApplicationCommand`/`ApplicationCommand` pairs of types (I think there are also others), which represent request/response data respectively.

That being said, I am not passionate about the specific name and am also happy to call it `SimpleAttachment`, `PartialAttachment`, or `RequestAttachment`, depending on whether we want to emphasize its nature or its role.

julmb: Oh, I guess the "metadata of existing attachments" docstring sounds a bit strange. The same `CreateAttachment` type is also used to represent the collected attachment metadata (both from existing attachments and new uploads). I updated to haddock documentation to hopefully make a little more sense.

src/Discord/Internal/Types/Channel.hs
julmb: Yes, the uniqueness issue still exists. In order to keep this PR as small as possible, I retained the existing behavior without introducing any unrelated changes.

The `createEmbed :: CreateEmbed -> Embed` function generates `attachment://` links to the embed attachments, so it would need to replicate the same indexing behavior. And this could still not prevent the case of a regular attachment (where the user gets to choose the filename) colliding with an embed attachment.

L0neGamer: Please can you add an issue for this in the repo, and mention the issue in a comment in this function, just so it isn't forgotten?

src/Discord/Internal/Types/Channel.hs
```diff
@@ -701,6 +714,72 @@ instance ToJSON Attachment where
...
+  prefix = T.filter (/= ' ') createEmbedTitle
+
+uploadsAssemble :: Maybe [CreateEmbed] -> [CreateUpload] -> [CreateUpload]
+uploadsAssemble embeds = (++) $ maybe [] (concatMap uploadsEmbed) embeds
```
L0neGamer: would it be simpler to just have `uploadsAssemble :: Maybe [CreateEmbed] -> [CreateUpload]` and let the other library code concat them together? might even make sense to drop the Maybe, but I think that's fine.

L0neGamer (commented):
Likely good to go, I want to test this out a bit myself but I ran the example locally and it all seemed to work. If I haven't gotten back to you by next sunday please do ping me.

julmb:
> Please can you add an issue for this in the repo, and mention the issue in a comment in this function, just so it isn't forgotten?

Done.

> would it be simpler to just have `uploadsAssemble :: Maybe [CreateEmbed] -> [CreateUpload]` and let the other library code concat them together? might even make sense to drop the Maybe, but I think that's fine.

Done. I do not remember why I originally made it `Maybe [CreateEmbed] -> [CreateUpload] -> [CreateUpload]`. I think it used to do something more complicated than it does now. Pushing the merging responsibility to the call sites seems reasonable. It also makes for a simpler name in `uploadsEmbeds` instead of `uploadsAssemble`.

> Likely good to go, I want to test this out a bit myself but I ran the example locally and it all seemed to work. If I haven't gotten back to you by next sunday please do ping me.

Sounds good to me. Do you still want to rename `CreateAttachment` to something like `SimpleAttachment`, `PartialAttachment`, or `RequestAttachment` or should it stay the way it is?

L0neGamer:
Sorry for the delay, would appreciate a rename to requestattachment, will merge soon after you do so thank you.

julmb:
> Sorry for the delay, would appreciate a rename to requestattachment, will merge soon after you do so thank you.

Done.

I also renamed `CreateUpload` to just `Upload`, since that name was still available and I think it's clearer and also more consistent that way. I'm not passionate about it tough, so if you don't like it I can easily reset the branch back to b3d4f53.

L0neGamer:
I was initially opposed to CreateUpload->Upload but I think it makes sense here!

@yutotakano would appreciate your eyes on this when you can, but I will merge this momentarily.

PR #247, opened 2026-05-21
Use Specific Type for `resolvedDataAttachments`

julmb:
According to the [documentation](https://docs.discord.com/developers/interactions/receiving-and-responding#interaction-object), this field is a "Map of Snowflakes to attachment objects", so we can parse this into a more specific type.

I tested it briefly and it seems to work for its intended use case of attachment parameters to chat input application commands.

### DIFF
--- a/src/Discord/Internal/Types/Guild.hs
+++ b/src/Discord/Internal/Types/Guild.hs
@@ -297,6 +297,17 @@ data Role =
       , roleMention :: Bool                      -- ^ Whether this role is mentionable
     } deriving (Show, Read, Eq, Ord)
 
+instance ToJSON Role where
+  toJSON Role {..} = object
+    [ "id" .= roleId
+    , "name" .= roleName
+    , "color" .= roleColor
+    , "hoist" .= roleHoist
+    , "position" .= rolePos
+    , "permissions" .= rolePerms
+    , "managed" .= roleManaged
+    , "mentionable" .= roleMention
+    ]
 instance FromJSON Role where
   parseJSON = withObject "Role" $ \o ->
     Role <$> o .: "id"
--- a/src/Discord/Internal/Types/Interactions.hs
+++ b/src/Discord/Internal/Types/Interactions.hs
@@ -35,6 +35,7 @@ where
 import Control.Applicative (Alternative ((<|>)))
 import Control.Monad (join)
 import Data.Default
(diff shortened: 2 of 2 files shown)


L0neGamer:
I think this could include user and role info without too much trouble, if you don't mind testing that as well?

julmb:
I tested each of the fields briefly just now:
- `resolvedDataUsers`
  - set when using `CreateApplicationCommandUser`
  - set when using `OptionValueUser`
  - parsed successfully as `Map Snowflake User` in simple example
- `resolvedDataMembers`
  - set when using `CreateApplicationCommandUser`
  - set when using `OptionValueUser`
  - cannot be parsed as `Map Snowflake GuildMember` because it uses [partial member](https://docs.discord.com/developers/interactions/receiving-and-responding#interaction-object-resolved-data-structure) values
  - parsed successfully as `Map Snowflake Value` in simple example
- `resolvedDataRoles`
  - set when using `OptionValueRole`
  - parsed successfully as `Map Snowflake Role` in simple example
  - does not work due to missing `ToJSON Role`, either that has to be added or `ToJSON ResolvedData` removed
- `resolvedDataChannels`
  - set when using `OptionValueChannel`
  - cannot be parsed as `Map Snowflake Channel` because it uses [partial channel](https://docs.discord.com/developers/interactions/receiving-and-responding#interaction-object-resolved-data-structure) values
  - parsed successfully as `Map Snowflake Value` in simple example
- `resolvedDataMessages`
  - set when using `CreateApplicationCommandMessage`
  - parsed successfully as `Map Snowflake Message` in simple example
  - documentation says this uses [partial message](https://docs.discord.com/developers/interactions/receiving-and-responding#interaction-object-resolved-data-structure) values but it worked with our `Message` type when I tested it
- `resolvedDataAttachments`
  - set when using `OptionValueAttachment`
  - parsed successfully as `Map Snowflake Attachment` over the last few weeks

I only really tested `resolvedDataAttachments` somewhat thoroughly since I actually used it in my own application. I can add the others, but I cannot give any guarantees.

julmb:
Added types for the other fields so you can take a look at it. Personally, I am only interested in the attachments one, so I'm happy to reset the PR to 2bebc86 if you want to play it safe.

L0neGamer (approved):
Your testing is sufficient for me, thanks for the work done!

